PR #82, opened 2021-02-22
Update dev_readme.md

gedankenstuecke:
The `dev_readme.md` looks good overall! I just added some code tags so that the filenames etc. stand out a bit more!


GeorgiaHCA:
Thanks for reviewing @gedankenstuecke ! I've tested it a couple of times now with people and it seems to be possible to follow. Happy to merge your changes as they are definitly an improvement and we can continue to make to update as more people join and test it! 🌟

PR #104, opened 2021-05-26
Lotty repo refresh

lottycoupat:
@sowmya9507 @GeorgiaHCA 

have added two folder:

- get-started - @sowmya9507 please could you add to this folder all the files you think are essentail to get set up for contributing to the platform development.
- platform-code - @sowmya9507 in the "home-page" is where I suggest you put your work for the landing page. 

@GeorgiaHCA what do you think about changing "Landing Page" to "Home page" - seems more intuitive?


sowmya9507:
@lottycoupat thank you, I will add all the files and also the code that I have worked on till now.

PR #109, opened 2021-06-03
Delete platform-design directory

lottycoupat:
Requirements docs moved over to Autistica repo.

Complete [PR 487](https://github.com/alan-turing-institute/AutisticaCitizenScience/pull/487) beforehand.


GeorgiaHCA:
I think it's great to have this in thee project management repository, but also important to document where issues and actions are coming from and why. For instance, issues for a piece of dev work on the interface should reference requirements/wireframe design. This can be done through links to the other repo, but I would check with @KirstieJane if it's best to stay here. 
(Unless the idea is that people will go to the other repo as a default to begin with in which case we need to figure out about issues)

lottycoupat:
@GeorgiaHCA The idea that I'm envisioning is that you would firstly be re-directed to the AutisticaCitizenScience repository first and anything would branch out from there. That was the thought process behind adding the platform-code folder so that anyone interested in coding for the platform can head to that GitHub straight away. Any other documentations should be held in the AutisticaCitizenScience repository unless it is step by step how to use the code and readmes. Within the Readme's we would be able to link out to the other repository if people need to find docs on wireframes/user flow/requirements. 

I'd be worried about having the same documents in both repos as that could confused people on where they should upload work they have done and what is being held in which repo?

What do we think?

GeorgiaHCA:
Closiing this PR as folder has now been updated

Issue #17, opened 2021-01-20
Create Sign-up/sign-in Flow


GeorgiaHCA:
### Summary

Create a flow for user to sign up to platform.

A wireframe MVP of the share experience page: 

[simplified sign up flow](https://www.figma.com/file/HxqTSdeyAUNMualnBV0PET/Autspace-(WIP)?node-id=168%3A1)

### Deadline 

16th Feb 2021

### What needs to be done?
- [ ] After a user successfully signs **up**, direct them to Frame 1
- [ ] Implement Frame 1
    - [ ] Add question "Are you Autistic?"
    - [ ] Add radio buttons with options: Yes, No, Not sure, Prefer not to say
    - [ ] Add `Continue` button that leads to Frame 4 
- [ ]  Implement Frame 4
    - [ ] Add question "If you have another connection to autism, please write it here. For example, sister, researcher, or friend. You can also leave this space blank."
    - [ ] Add radio buttons with options listed in the image below.
    ![CleanShot 2021-06-29 at 07 56 51](https://user-images.githubusercontent.com/1872836/123793336-95db8480-d8af-11eb-8374-b045a3b78850.jpg)

    - [ ] Add `Submit` button that directs user to Confirmation navigation page (if #121 has been implemented) or to the landing page if #121 has not been implemented.  

### Archived 
Leaving this tasks here for posterity. 

- Allow user to create username and password to sign-in to platform
- Must use Open Humans sign in process: see diagram in the [user-flow doc](/user-flow.md)
- Must seek to collect user's autism status (but not mandated requirement) with option to self-describe
- Must allow users to sign in as moderators (links to #100 )

### Who can help

* Anyone with web development skills to build
* Anyone from the community to review and test*
* the designs are already founded on substantial community feedback and user testing

GeorgiaHCA:
Are you able to take this on @sowmya9507 ?

GeorgiaHCA:
<img width="1930" alt="image" src="https://user-images.githubusercontent.com/46889966/119357176-2a463c00-bc9f-11eb-9319-26938bd6c2fc.png">

GeorgiaHCA:
Original user testing feedback

From https://github.com/alan-turing-institute/AutisticaCitizenScience/blob/master/community-recommendations/user-tests/sprint-6-feedback/summary-sprint6-feedback.md

‘How it works’ section

    A C d: “I want to look only at ‘how it works’.”
[...]

PR #122, opened 2021-07-01
Sign up userflow

mridubhatnagar:
Pull request for SignUp page - https://github.com/alan-turing-institute/AutSPACEs/issues/17

Sign Up Page Preview

Screenshot from 2021-06-24 00-56-32 (1)


jhlink:
Wonderful work, @mridubhatnagar !! You're missing Frame 4, but after implementing that screen, this PR should be good to merge!!

mridubhatnagar:
@jhlink I have added Frame 4 implementation in the PR. 
The links and buttons at the moment are not functional. I wasn't sure about the underlying URLs for each of them.

Screenshot from 2021-07-05 13-49-19

server/apps/main/templates/main/signup.html
```diff
@@ -0,0 +1,125 @@
...
+            </li>
+            <li class="nav-item">
+                <a class="nav-link" href="#">LOGIN</a>
+            </li>
+          </ul>
+    </nav>
+    <div id="project-sign-up">
+        <p>Sign Up to AutSPACE</p>
+    </div>
+    <form id="form-check" action="{% url 'signup_frame4_test' %}" method="POST">
```
jhlink, line 102: You're missing `main:` before `signup_frame4_test`. 

Otherwise, I get an error during template rendering. 
CleanShot 2021-07-07 at 12 49 37

server/apps/main/templates/main/signup.html
```diff
@@ -0,0 +1,125 @@
...
+                <input type="radio" name="inlineRadioOptions" id="inlineRadio1" value="option1">
+                <label for="explicit-label-name">NO </label>
+                <input type="radio" name="inlineRadioOptions" id="inlineRadio2" value="option2">
+                <label for="explicit-label-name">NOT SURE </label>
+                <input type="radio" name="inlineRadioOptions" id="inlineRadio3" value="option3">
+                <label for="explicit-label-name">PREFER NOT TO SAY </label>
+                <input type="radio" name="inlineRadioOptions" id="inlineRadio4" value="option4">
+            </div>
+            <div class="submit">
+                <button type="button" class="btn btn-primary mb1 bg-greenyellow" form="form-check">CONTINUE</button>
```
jhlink, line 118: Clicking the submit button doesn't direct me to the next page. I'm not sure if the `form action` html attribute is supposed to create that effect, but you can sort of _cheat_ by surrounding `CONTINUE` with 
```html
<a href="/main/signup1">[TEXT]</a>
``` 

Of course, this won't send the form data to the server, but at the moment, it's not a concern for this current milestone. We're focused on establishing a flow, even if it's roundabout.

jhlink (changes requested):
You're so very close @mridubhatnagar !! Just a couple more changes!!

server/apps/main/templates/main/signup.html
mridubhatnagar: Yes, I was facing the same issue. 

I have pushed the changes and updated it using the anchor tag thing. The caveat is that even without selecting and submitting the choices it user would go onto the next page by clicking on the submit button.

server/apps/main/templates/main/signup.html
```diff
@@ -115,7 +115,7 @@
                 <input type="radio" name="inlineRadioOptions" id="inlineRadio4" value="option4">
             </div>
             <div class="submit">
-                <button type="button" class="btn btn-primary mb1 bg-greenyellow" form="form-check">CONTINUE</button>
+                <a href="/server/apps/main/templates/main/signup1.html"><button type="button" class="btn btn-primary mb1 bg-greenyellow" form="form-check">CONTINUE</button></a>
```
jhlink, line 118: Hey @mridubhatnagar , so for this href, you don't provide the absolute path to the html file on the server. Rather, you provide the _URL_ path/pattern that directs to the `signup1.html` page. 

For example, the URL pattern that directs users, according to the `urls.py` that you've modified is `signup1`. The django app, which is what we're developing, is called `main`. 

So the `href` link would now become `/main/signup1`. 

Currently, if I try to navigate to the second page, I get this 404 error .
CleanShot 2021-07-08 at 07 45 07

jhlink (changes requested):
Just one last change!! :)

server/apps/main/templates/main/signup.html
mridubhatnagar: Hey @jhlink 

Yes, makes sense. 

So, I tried replacing the link to /main/signup1/ and tested the implementation on my local. The redirection to the next page doesn't work. I get an error saying 

```
Cannot GET /main/signup1
```

I then modified the URL to Django template-based URL that we earlier discussed. 

```
<div class="submit">
                <a href="{% url 'main:signup_frame4_test' %}"><button type="button" class="btn btn-primary mb1 bg-greenyellow" form="form-check">CONTINUE</button></a>
</div>
```
On using the above URL I get a bad request error.

jhlink: Awesome. This works great!!

jhlink:
Changes are great!! Merging.

jhlink (approved):

PR #132, opened 2021-07-18
removed comments in .yml files causing a docker compose error

badgermind:
see https://github.com/docker/compose-cli/issues/1896 to fix issue with Docker Compose version v2.0.0-beta.6


jhlink:
Understood. Merging. 

Thanks for the change, @badgermind !!

PR #160, opened 2021-08-19
[WIP] - Implement share experiences

mridubhatnagar:
This pull request is currently work in progress for issue ref https://github.com/alan-turing-institute/AutSPACEs/issues/156.

PR #187, opened 2021-10-25
Fix: Clean up database migrations

jhlink:
## Description
Cleaning up the database schema to facilitate #160 . 

When attempting to migrate to a new table, database migrations issues have occurred. 

Warning: When merging this PR, developers may have to reset their database. 

### Test
After pulling the PR and running the site, verify that `View Experiences` loads properly. 
If the webpage fails to loads with an error that a database relation does not exist, your database will need to be reset. 

#### Resetting the database
1. While docker is running, run bash with the following command. 
`docker container exec -it [container_name] bash`
2. Run the following command in the shell. 
`./manage.py sqlflush | ./manage.py dbshell`

#### Explanation
`sqlflush` is a command provided in `python manage.py` that generates all the SQL statements needed to reset the database to a clean state. 
`|` is a pipe command which passes the stdout output of the previous command to the input of the following command. 
`dbshell` is a command provided in `python manage.py` which allows the user to interact directly with the PostgresSQL `AutSPACEs` database.


Islast (approved):
Looks great to merge to me, assuming there aren't any databases currently running on the old schema that we need to worry about

PR #201, opened 2021-11-02
fix: Replace sh with bash in entrypoint

jhlink:
## Description
When running docker in a Windows environment, given that the user is using VSCode as the IDE and installed Desktop Git and Desktop Docker, executing `./docker-run.sh` fails with the following error message. 

`/usr/bin/env sh Permission denied`

## Definition of Done
- [x] Apply fix
- [ ] Add problem and solution to troubleshooting doc. 

## The Fix
Use `bash`, which is globally available, instead of `/usr/bin/env sh`, which has issues with execution. 

### Additionals
If this above doesn't work, here are a few additional steps to attempt. 

1. Within VSCode, ensure the end of the line sequence is set to `LF` not `CRLF`
CleanShot 2021-11-02 at 07 55 57
2. Add the path to `C:\Program/ Files\Git\bin\` to the `PATH` environment variable.
3. Verify that Docker is running.

### DIFF
--- a/docker/django/entrypoint.sh
+++ b/docker/django/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/bin/bash
 
 set -o errexit
 set -o nounset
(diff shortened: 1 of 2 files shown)


GeorgiaHCA:
Would anyone with a Windows computer be able to test this? 🙏 
@Angelsaur perhaps you could try?

PR #214, opened 2022-02-22
Frontend implementation of the platform 

soledadli:


BrainonSilicon:
Hi @soledadli! Awesome work on the platform's design!!! It's so cool to see it implemented. 

I'll go through this PR in the next few days and add any review comments that would be helpful.

GeorgiaHCA:
This is truly fantastic work @soledadli ! I've been looking at the deployed version and the design is awesome. There are a few things I've noticed around the text content and flow which need changing but I think this makes sense to add as separate issues and fix forwards so we can get this merged to the main repo!

GeorgiaHCA:
Are you able to take a look @BrainonSilicon? Do you reckon we can go ahead and merge this now?

callummole:
@helendduncan and I reviewed most of this today. I think the design is really nice. There are obviously bits that we would want to change (e.g. hardcoded experiences, some links here and there not working...), but I think we should merge this to main now, and branch off from there, since the whole site is still in dev. This PR makes a big positive change to the look and feel of the site, well done @soledadli!

callummole:
@GeorgiaHCA @BrainonSilicon do we have a practice for merging? I'd be happy to just go ahead and merge but want to check how you work first.

PR #394, opened 2022-12-08
Functioning `my stories`.

### LINKED ISSUE #153: Render a User's Personal Experiences in the `My Stories` Page
### Summary

When a user creates a personal experience on the site, it should get displayed in the `My Stories` page. This is where a user can - 
* See all submitted experiences in one page
* Delete any experiences 
* Edit any experience

The current goal of this issue is to render those experiences from the Open Humans database to the my_stories page and ensure that the page has the above 3 functionalities.

### What needs to be done?
**The page should only be viewed in the logged in state.**

- [x]  Shows an experience in row form
    - [x] Title of experience
    - [x] Creation date of the experience
    - [x] Description of the experience
    - [x] Show Public/Not Public option (Would be shown as a checkmark/no checkmark under the public column of the experience)
    - [x] Show Research/Not Research option (Would be shown as a checkmark/no checkmark under the research column of the experience)
    - [ ] Show Moderation Review Status 
         * Moderation status has four states: Draft, In Review, Approved, Rejected

Each row should have a delete and edit button. 
- [x] Delete button
       * Clicking the ‘delete’ button, the user should be able to delete the experience.
- [x] Edit button. Clicking the ‘Edit’ button, the user should be directed to a `Share Experience` page with all input fields/check boxes populated with the clicked experience data. 
- [ ] Clicking the `Submit` in `Share Experience` will update the experience with the new data. 
- [ ]  See the moderation status of each experience
       *  The status would be visible on each experience(possibly in a box in the corner of the experience )

### Who can help?

The dev team.

callummole:
Issue #153 gives a well-described issue about what is needed for viewing and managed your own stories. The required tasks are copied below. The description of what has been done in this PR is below the quote.

> ### What needs to be done?
> **The page should only be viewed in the logged in state.**
>
> - [x]  Shows an experience in row form
>     - [x] Title of experience
>     - [x] Creation date of the experience
>     - [x] Description of the experience
>     - [x] Show Public/Not Public option (Would be shown as a checkmark/no checkmark under the public column of the experience)
>     - [x] Show Research/Not Research option (Would be shown as a checkmark/no checkmark under the research column of the experience)
>     - [ ] Show Moderation Review Status 
>          * Moderation status has four states: Draft, In Review, Approved, Rejected
> 
> Each row should have a delete and edit button. 
> - [x] Delete button
>        * Clicking the ‘delete’ button, the user should be able to delete the experience.
> - [x] Edit button. Clicking the ‘Edit’ button, the user should be directed to a `Share Experience` page with all input fields/check boxes populated with the clicked experience data. 
> - [x] Clicking the `Submit` in `Share Experience` will update the experience with the new data. 
> - [ ]  See the moderation status of each experience
>        *  The status would be visible on each experience(possibly in a box in the corner of the experience )
> 

The `Share_Experience.html` page now uses a `ShareExperienceForm` to control data upload. This form is also used when Editing shared experiences. 

1. **Page view in logged in state:**  *My Stories dropdown menu unavailable when logged out- direct linking to the my_stories page redirects to the landing page* 
2. **Shows an experience in row form:**  *Title, and creation date of experiences are processed and sent to open humans, then read-back when `my stories` list is populated with default responses if key not available. A description of the triggering labels and public sharing options are sent as tags to OH and used to populate the rows entitled "triggering labels" and "Tags". Here explicit labels rather than checkboxes are used.*
3. **Each row should have a delete and edit button:** *Rows contain edit, download, and delete buttons. When delete is chosen, the file is deleted from OH. When edit is chosen the share experience page is populated with the data from the corresponding OH file. On submit the original file is deleted and the edited file (with corresponding new creation date even if nothing else is changed) is produced.*

**Question:** Does the information now available for each of `my stories` satisfy the requirement that there is a description of the experience in row form?

**Question:** When calling all OH files (my_stories) - are we calling all OH files associated with this project or all of the users OH files. Do we need to apply some filtering?

**Ongoing:** Moderation. There are some underlying decisions that need to be made about how and by who moderation is performed so it is suggested that the moderation tasks are migrated to a separate issue. Currently there is a column dedicated to moderation status which now contains only placeholder text.

This PR closes #153. We will open a moderation specific issue since that also needs admin roles.


gedankenstuecke:
Thanks for the nudge @helendduncan! I made a note to look over this on Monday!

server/apps/main/forms.py
```diff
@@ -0,0 +1,52 @@
...
+    abuse.group = 2
+    violence = forms.BooleanField(label = 'Violence and Assualt', required=False)
+    violence.group = 2
+    drug = forms.BooleanField(label = 'Drug and/or Alcohol misuse', required=False)
+    drug.group = 2
+    mentalhealth = forms.BooleanField(label = 'Mental Health Issues', required=False)
+    mentalhealth.group = 2
+    negbody = forms.BooleanField(label = 'Negative body image', required=False)
+    negbody.group = 2
+    other = forms.CharField(label='',
```
gedankenstuecke, line 35: Screenshot 2022-12-12 at 10 03 18

I noticed is that the "other" category for data entering in the form doesn't have a checkbox attached to and I think it might be a result of how it's set up as a form field here. It's not a bit problem, but I think standard UI convention might be to have the checkbox regardless?

server/apps/main/models.py
```diff
@@ -5,6 +5,7 @@
 class PublicExperience(models.Model):
     experience_text = models.TextField()
     difference_text = models.TextField(default="")
+    title_text = models.TextField(default="")
```
gedankenstuecke, line 8: This is maybe a broader design question, but I wonder whether this field should be required rather than set to an empty default? As the *My Stories* and other UI parts focus heavily on the presence of a title it might be better to allow for that? @GeorgiaHCA do you know if a decision was made on that at some point?

server/apps/main/urls.py
```diff
@@ -14,7 +14,7 @@
         name='view_experiences'),
     url(r'^moderate_public_experiences/?$', views.moderate_public_experiences,
         name='moderate_public_experiences'),
-    url(r'^upload/?$', views.upload, name='upload'),
```
gedankenstuecke, line 17: As a head's up, this change breaks the currently existing `/main/list` view. But that view might be removed in any case in favor of this PR if I understand it correctly.

gedankenstuecke (commented):
Really great work @helendduncan & @callummole, I just ran it locally and it works really well! 🎉🎉🎉 I also looked over the code and just had a few minor comments here and there. 

Regarding your questions: 

> Question: Does the information now available for each of my stories satisfy the requirement that there is a description of the experience in row form?

I would say so, it looks very clean and from what I can tell the current version also bypasses the need to actually read any of the files on Open Humans, which is great in terms of speed. If we make the creation of a title for each experience mandatory then this should work really well! 

> Question: When calling all OH files (my_stories) - are we calling all OH files associated with this project or all of the users OH files. Do we need to apply some filtering?

I don't think any filtering should be necessary for the actual production environment, as AutSPACEs won't request access to non-AutSPACEs files and there won't be any non-experience files within AutSPACEs. As a side-note: As I re-used one of my own development/test OH projects the list was populated by a lot of files that weren't associated coming from AutSPACEs, but even that didn't break anything! 

> Ongoing: Moderation. There are some underlying decisions that need to be made about how and by who moderation is performed so it is suggested that the moderation tasks are migrated to a separate issue. Currently there is a column dedicated to moderation status which now contains only placeholder text.

That sounds very reasonable to me! 👍 

And again, really well done!

callummole:
> Question: When calling all OH files (my_stories) - are we calling all OH files associated with this project or all of the users OH files. Do we need to apply some filtering?

@helenduncan No. We are working with the specific user's openhumansmember object here, so we are only listing files for them.

server/apps/main/urls.py
callummole: @gedankenstuecke correct. There was some redundancy in the code. There were some pages that never actually were loaded. (`list.html` included). We adapted some code from the list view/html pages for the new `my_stories`.

It seems that there were some efforts to abstract the front-end into a central 'application' (`application.html`), which was to be extended in sub-pages (e.g. for `list.html`). I'm not sure about the chronology here, because they weren't actually used. 

I was going to keen the redundant pages/views for now and tidy things up when we are closer to an MVP since we found code in the unused pages to be useful.

callummole:
> I would say so, it looks very clean and from what I can tell the current version also bypasses the need to actually read any of the files on Open Humans, which is great in terms of speed. If we make the creation of a title for each experience mandatory then this should work really well!

Currently this comes free with using the forms approach. The title, experience, and suggestions are currently required, and one will not be able to submit an experience without them. I assume that we want the suggestions to be optional?

server/apps/main/models.py
callummole: @gedankenstuecke it is actually required on the forms side in `share_experience.html`, which is AFAIK the only method of creating a story. So even though it is not required on the model view I think that users cannot create experiences without titles (or, currently, experiences or suggestions, see comment in PR issue). 

If you are dealing with the database locally you can still create entries without a title though.

server/apps/main/urls.py
gedankenstuecke: Got it, just wanted to flag it in case it's a real issue. But then we can ignore this for now and leave the clean up of unused things for closer to the MVP

server/apps/main/forms.py
gedankenstuecke: I *think* the standard way of handling it is to have a checkbox regardless with a label *Other* and then have the free text box afterwards (see the image below for how Google Forms does it). 

Screenshot 2022-12-12 at 13 05 38

I think this *might* be more accessible, as there is a clear label for the checkbox and text field if one does it like this. If you look at my screenshot above it becomes unclear what the text box relates to once you start typing as the 'example' text disppears, instead it then looks like the text box is related to mental health issues.

callummole: Yes. I see exactly what you mean.

server/apps/main/models.py
callummole: I have moved this to a discussion #400, because it is not currently actionable (since that is the criteria we discussed :) ).

gedankenstuecke: Yep, that makes sense!

PR #397, opened 2022-12-08
Populate triggering labels column

helendduncan:
Tags for triggering labels added. Shown in appropriate columns for viewing my_stories.

Need to decide behaviour for "Other"


callummole (approved):
nice. I have made some minor additions:

- removed the links given to the sharing tags. I think a cleaner work flow is to force all edits to happen through the `Edit` button. 

- used our custom tag `define` to specify a sharing_tag list rather than repeating ourselves.

- I have also shortened the labels.

PR #402, opened 2022-12-15
add other label and left alignment

callummole:
In [this discussion](https://github.com/alan-turing-institute/AutSPACEs/pull/394#discussion_r1045807012) it was suggested to keep a checkbox alongside `Other`, since when you type in `Other` the placeholder disappears.

The simplest solution is to attach an `Other` label, which I have done in this commit. Keeping the checkbox and textbox in sync (e.g. automatic typing enables the checkbox) requires Javascript. Though this is probably simple enough to implement, it would take longer than the few mins dedicated to this solution. I thought it was good enough for now, in order to prioritise other developments towards the MVP.

### DIFF
--- a/server/apps/main/forms.py
+++ b/server/apps/main/forms.py
@@ -32,9 +32,11 @@ class ShareExperienceForm(forms.Form):
     mentalhealth.group = 2
     negbody = forms.BooleanField(label = 'Negative body image', required=False)
     negbody.group = 2
-    other = forms.CharField(label='',
+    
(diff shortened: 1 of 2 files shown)


gedankenstuecke (approved):
I think that's a reasonable solution for now, using javascript should hopefully see straightforward in the future, [see stackoverflow here](https://stackoverflow.com/questions/14556692/check-checkbox-if-inputbox-has-content).

PR #410, opened 2023-01-05
Delete PublicExperiences at the same time as OH

### LINKED ISSUE #406: Delete PublicExperience object on (local) database when deleted on OH
### Summary

Currently when a user deletes a story from the OH database via the AutSPACEs site - the local PE object containing the contents of the story is not deleted

### Deadline 

Feb 2023

### What needs to be done?

- [x] In functions where `request.user.openhumansmember.delete_single_file` is called, delete the corresponding PE object at the same time

### Who can help?

* @helendduncan 
* @callummole 

---

### LINKED ISSUE #407: Add confirmation warning for deletion of files
### Summary

Add a confirmation/warning step when a user is about to delete OH files

### Deadline 

Feb 2023

### What needs to be done?

- [ ] Enact a warning (?page?) when `request.user.openhumansmember.delete_single_file` is called

### Who can help?

* @callummole 
* @helendduncan 

---

callummole:
Changes made:

Closes #406 
- add a hidden `uuid` field to a form that will get populated if the file is editable.
- create a `delete_single_file` function that also deletes from the PublicExperiences database if given a uuid.
- update PublicExperiences if edited an existing viewable experience; delete if changed to non-viewable.
- use `uuid` as filename for openhumans, so that we can delete using `uuid`.
- remove unnecessary `file_id` form field.

To do to Close #407 
- [x] redirect delete button to a view that calls the `delete_single_file` function.
- [x] add warning.


gedankenstuecke (changes requested):
Thanks for taking a stab at this @callummole! I fear we might need to think a bit harder about the approach for how implement editing/updating experiences in this way to avoid potential security issues: 

I'll try to outline the process as implemented to see if I understood it correctly: 

Currently there is one form that is used for both submitting and editing experiences and the form contains two hidden values for the `file_id` (i.e. the Open Humans file ID) and the `uuid` (our internal ID for public experiences).

If it is an edit of an existing experience it will call the `delete_single_file` function which takes both the `file_id` and the `UUID` parameter to delete both the file on OH as well as the public experience. 

If this summary is correct, then I see two potential issues with this: 

## Validation 
This process would need some validation as it otherwise is a security problem. The `hidden` form fields might be hidden from the UI, but are accessible in the browser and the data submitted can quite easily be modified. For the `file_id` of Open Humans that is less of an issue, as you can't delete files that aren't associated to the `request.user.openhumansmember`. 

But it's a bigger issue for the `PublicExperience`, as the `delete_single_file` function currently does not check if the `uuid` actually belongs to the user. So someone could modify the hidden `uuid` form field to the `uuid` of an experience of another user, thus leading to that other users experience being deleted or edited.

## Deleting or updating public experiences
This is rather minor, but: Should edits to an experience that has a public experience really delete the public experience? Or should the public experience be updated and maybe flagged as not-yet-moderated? 


## Potential solutions?

An easy fix for the validation would be to just double check that the experience of the UUID actually belongs to the `request.user`. 

But while writing this up I also wondered if we actually need both the `file_id` and `uuid` in the actual form at all. I _think_ each individual experience, whether also existing as a `PublicExperience` or not should have the `uuid` as part of its metadata in OH, right? In that case we could get rid of the `file_id` in the form and just submit the `uuid` as it will allow us to delete the correct file on OH and also find any public experience. 

The one downside to this approach would be that we'd need to do another `.list_files()` call to Open Humans to get the right file ID after getting the form is submitted, but it would allow us to verify that this user actually has the UUID in their account. But maybe I'm overlooking some other drawbacks that this approach would have? 

Also happy to jump on a call or sit down for a chat in person (will be in tomorrow) in case that helps!

callummole:
@gedankenstuecke there is now a `update_public_experience_db()` function that controls whether to update, insert, or delete, depending on viewable status. 

I agree that it's cleaner logic.

gedankenstuecke:
Yeah, wanting to speed up actions by avoiding API calls makes total sense. I wonder if the following could be a potential workaround that actually just occurred to me now: 

Open Humans has the function to delete files based on their filename (`oh_member.delete_single_file(file_basename='FILENAME')`). 

Currently we use a more verbose and more human-readable filename: https://github.com/alan-turing-institute/AutSPACEs/blob/e5a4890b5b401b0925192b250510c2d1c4b1721e/server/apps/main/views.py#L106 

This does not easily lend itself for using the filename for the deletion (especially if people were to edit their experience's titles etc). 

But, if we were willing to sacrifice the filename-readability for something like `99b518b2-8dcd-11ed-8f87-6afd71b9c647.json`, i.e. using the AutSPACEs `UUID` as the filename, then we could delete the files on the OH site without having to query for a OH-file-ID or having to explicitly pass along. 

My own feeling is that this might be an acceptable trade-off, as the filenames are not shown anywhere on the AutSPACEs side of things and will be invisible to 99% of all users. The only place they are visible to users are on the Open Humans activity page (`https://www.openhumans.org/activity/ACTIVITY_NAME/`), which most people will never really look at. And even if, on those pages the `description` metadata is shown as well, which should help users understand what's in the file and can provide more information than any filename would. 

How does that sound?

callummole:
@gedankenstuecke that sounds like a clever solution!

callummole:
@gedankenstuecke that works well. I'm happy to merge, if you are?

server/apps/main/templates/main/deletion_success.html
```diff
@@ -0,0 +1,72 @@
...
+  <title>View Stories</title>
+</head>
+<body>
+{% include 'main/partials/navigation.html' %}
+
+<section class="confirmation">
+  <div class="confirmation-container">
+    <div class="confirm_success">
+      <h1 class="big-heading">Success!</h1>
+      <p class="title-text">Experience '{{title}}' does not exist for the current user</p>
```
gedankenstuecke, line 31: This wording is a bit confusing to me, when I played around with it and encountered the `does not exist for the current user` wording I assumed I ran into an error, rather than success. But I think this is always displayed on deletion of a public experience?

server/apps/main/views.py
```diff
@@ -49,44 +49,95 @@ def logout_user(request):
...
+            difference_text=data['wish_different'],
+            title_text=data['title'],
+            open_humans_member=ohmember,
+            experience_id=uuid,
+            abuse=data['abuse'],
+            violence=data['violence'],
+            drug=data['drug'],
+            mentalhealth=data['mentalhealth'],
+            negbody=data['negbody'],
+            other=True if data['other'] != '' else False,
```
gedankenstuecke, line 117: This confuses me slightly, does it store the actual text of the 'other' label or just a binary True/False?

gedankenstuecke (approved):
Just finished looking over it and also testing it locally and that seems to all work as expected now! I just added two minor comments where some clarification would be good!

callummole:
Merging for now to keep branch short-lived. @gedankenstuecke's comments above relate to issues about 1) rewriting to use `GET` methods instead of unnecessary `POST`s, and 2) align `PublicExperiences` with form fields. (add links...)

server/apps/main/templates/main/deletion_success.html
gedankenstuecke: This will also be solved by #411

PR #413, opened 2023-01-12
Remove unnecessary POST request to GET request

### LINKED ISSUE #411: Change GET & POST routes
### Summary

Currently there is some slightly inconsistent use of `PUT` and `POST` methods when viewing/editing data. This leads to a lot of data needing to be passed around as hidden form fields where not really necessary. 

Broadly speaking: `GET` requests are considered idempotent and should be used for any requests that purely request data, in our case that would e.g. be fetching someones experiences and displaying a form for submitting/editing an experience.

`POST` requests on the other hand should be used for things that actively submit and change data, e.g. submitting a form for a new/edited experience or deleting an experience. (Also see [the comparison of `GET` and `POST` here](https://www.w3schools.com/tags/ref_httpmethods.asp)). 

This is a follow-up from a discussion around #410 

### What needs to be done

I would suggest changing some of our routes with the `GET` / `POST` paradigm in mind: 

- Change the viewing of the `edit` forms to `GET` requests. For creating new experiences this form would be empty, for editing they would be prefilled by directly fetching the experience data from the database, through a URL parameter (e.g. `/experience/edit/{{uuid}}`). 
- Change the pre-confirmation page for deleting experiences to `GET` (similarly, through `/experience/delete/{{uuid}}`). 
- Only keep the actual submission of the form for editing/deleting as `POST`
- The `My Stories` tab can be streamlined after doing this, as not all experiences need to be included as hidden form field content.

callummole:
Closes #411.

A few changes limit the amount of POST requests:

Url parameters:
- `delete_exp` takes a uuid and title. 
- `edit_exp` takes a uuid. 
     - There is a new helper function, `get_oh_file()`, that, given a uuid, retrieves the file data.

`delete_exp` and `edit_exp` are called as GET requests from the `my_stories` tab (see description in #411).

`POST` requests are only executed when data is to be changed:
- on deletion.
- on submission (of new or edited experience). 

I have also deleted the now-redundant hidden form field `uuid`.

Comments for the future:
- the url passed to `delete_exp` could be very long since it includes the title. It might be nice to do the confirmation on the same page, with javascript, rather than have separate confirmation and success pages.

### DIFF
--- a/server/apps/main/forms.py
+++ b/server/apps/main/forms.py
@@ -39,17 +39,12 @@ class ShareExperienceForm(forms.Form):
                             widget=forms.TextInput(), required=False)
     other.group = 2
     
-    
     # sharing options
     viewable = forms.BooleanField(label = "Share on AutSPACE website", required=False)
     viewable.group = 3
     research = forms.BooleanField(label = "Share for research", required=False)
     research.group = 3
-    
-    
-    # hidden field that tracks PublicExperience uuid (which is also the OH filename) when you're editing an experience
-    uuid = forms.CharField(required=False, widget=forms.HiddenInput())
-    uuid.group = "hidden"
+
     # hidden field for moderation status
     moderation_status = forms.BooleanField(widget = forms.HiddenInput(), required=False, initial=False)
     moderation_status.group = "hidden"
--- a/server/apps/main/templates/main/deletion_confirmation.html
+++ b/server/apps/main/templates/main/deletion_confirmation.html
@@ -36,9 +36,7 @@
     
     <div class="row mx-0 py-4 justify-content-start">
       <div class="col-2">
-            <form action="{% url 'main:delete_experience' True %}" method="post">{% csrf_token %}
-              <input type="hidden" name="uuid" value="{{uuid}}"/>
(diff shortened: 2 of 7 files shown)


gedankenstuecke (approved):
This works well as a hot fix for the problem!

gedankenstuecke:
@callummole I could see two potential routes for the future to improve upon this & remove the need for passing the title: 

1. Using JavaScript we could implement [a modal view](https://getbootstrap.com/docs/5.3/components/modal/) that pops up with a confirmation to delete the page in question. That modal could then have the actual `POST` for the deletion of the experience. 
2. If we want to keep the separate page, we could use the `sessions` option in Django. In the view that creates the _My Stories_ page we could set a cookie that has all UUID/Title associations, and those could then be accessed in the view that renders the deletion confirmation. See https://docs.djangoproject.com/en/4.1/topics/http/sessions/ for more details on that option. 

I don't have a real preference for either of those to be honest and we can see both with community members what they would prefer as an improvement (keep the extra confirmation page vs modal view).

helendduncan (approved):

callummole:
Thanks both. I've made a note of your suggestions @gedankenstuecke for version 2!

Issue #162, opened 2021-08-24
Implement Moderation page


jhlink:
## Description
When experiences are submitted for review, a moderator will need to review the submission for adherence to community conduct and so forth.  

### Tasks
- [x] Populate a table of experiences, tagged with "review"
- [ ] Add an approve button that sets the moderation state of the experience to "Accepted/Approved/Public"
- [ ] Add a reject button that sets the moderation state of the experience to "Rejected"

SuziQpid:
Suggest another button is added after rejection that might read ' further action required' ' contact user to request edit/ ' or some other variation

jhlink:
@SuziQpid , certainly. There are additional features/requirements to add from a user perspective if an experience is rejected, such as a change log or even some sort of reason for rejection. The moderation flow between moderators and users has yet to be hashed out fully, so for the time being, a `reject and try again` approach will be implemented for simplicity and progress sake.

PR #414, opened 2023-01-16
Implement moderation page

helendduncan:
PR to address tasks raised in #162. Copied below for information:

> ### Tasks
> - [x] Populate a table of experiences, tagged with "review"
> - [x] Add an approve button that sets the moderation state of the experience to "Accepted/Approved/Public"
> - [x] Add a reject button that sets the moderation state of the experience to "Rejected"

moderate_public_experiences has been updated to receive both un-moderated and moderated experiences from the (local) database of PublicExperience objects.

Moderators can access either by clicking on the accordion link and getting a list of the relevant experiences.

The "moderate new experience" or "review moderated experience" buttons take the moderators back to the shared_experiences form as a user would see if they were editing their own experience. With the addition of a moderation drop down menu to change the status.

Additional follow on work separated into issue #418

### DIFF
--- a/server/apps/main/forms.py
+++ b/server/apps/main/forms.py
@@ -46,6 +46,20 @@ class ShareExperienceForm(forms.Form):
     research.group = 3
 
     # hidden field for moderation status
-    moderation_status = forms.BooleanField(widget = forms.HiddenInput(), required=False, initial=False)
+    statuses = [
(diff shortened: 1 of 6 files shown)


callummole:
Closing since basic moderation page is in place. Work continuing on #418

Issue #166, opened 2021-09-07
Add user role assignment via Django Admin


jhlink:
## Description
User roles/permissions will be managed through the Django admin panel. An admin by default will be able to view, manipulate, and remove all data and users. However, the primary utility for feature will be assigning which users are moderators. 

### Tasks
Users use their OpenHumans credentials to utilize the site. 
- [x] Create a new or use the pre-existing users table, provided in the Django admin library, and create a foreign key on the OpenHumans user ID
- [x] Create a Moderator status
- [x] Allow admin to set a user as moderator
- [x] Allow admin to set a moderator as a user
- [ ] Add documentation on how to create an admin user for development

SuziQpid:
This seems very straight forward. Congratulations. If l can get my head around this, most people will.

anoura12:
@jhlink would the modboss/reviewer role be added to this issue as well or will that be a separate issue altogether?

jhlink:
@anoura12 , please create a separate issue for this.

callummole:
Closed in #420

PR #420, opened 2023-01-23
Set up Admin pages and permissions

helendduncan:
Addressing issue #166 copied below

> ## Description
> User roles/permissions will be managed through the Django admin panel. An admin by default will be able to view, manipulate, and remove all data and users. However, the primary utility for feature will be assigning which users are moderators. 
> 
> ### Tasks
> Users use their OpenHumans credentials to utilize the site. 
> - [x] Create a new or use the pre-existing users table, provided in the Django admin library, and create a foreign key on the OpenHumans user ID
> - [x] Create a Moderator status
> - [x] Allow admin to set a user as moderator
> - [x] Allow admin to set a moderator as a user
> - [ ] Add documentation on how to create an admin user for development

Custom template tag to limit permissions based on user status (see below)

### DIFF
--- a/server/apps/main/templates/main/partials/navigation.html
+++ b/server/apps/main/templates/main/partials/navigation.html
@@ -1,4 +1,6 @@
  {% load static %}
+ {% load custom_tags %}
+
 <!--    Stylesheet-->
   <link href="{% static '/css/navigation_light.css' %}" rel="stylesheet" id="mystylesheet">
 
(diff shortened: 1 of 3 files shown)


callummole:
The custom tag will be something like:

```
@register.simple_tag
def toggle_story(user):
  return Group.getobject(user=user).name = "Moderators:
```

callummole:
@helendduncan I added the tag and toggled the link according to group membership. I don't have time to finish up now, but remaining is:
- [ ] add checks on whatever moderation pages there are, so that no one can hard-code the url.

- [ ] currently the site fails if you are an admin trying to navigate the site, because you are not an `openhumansmember`. Not a huge problem because admins will deal with the admin site. But we should add in checks for this when we check if a user is authenticated and adjust behaviour accordingly (which could mean hard-coded stories). 
- [ ] also need to add brief documentation on how to create a superuser, a `Moderators` group, and how to add a user to a group, for development.

gedankenstuecke:
Regarding the second part for the admin/superuser not being an OH member: One could solve this by not using the main `admin` account but rather use "real" existing user accounts which do have the associated OH object?

callummole:
> Regarding the second part for the admin/superuser not being an OH member: One could solve this by not using the main `admin` account but rather use "real" existing user accounts which do have the associated OH object?

Indeed, you can make a quick 'admin-only' superuser to get you started, then use that to assign existing users `staff` or `superuser` status. This works fine.

gedankenstuecke:
Yep, that's what we did with Open Humans itself as well :D

callummole:
will fold documentation into a dedicated issue, since there is a bunch to make and we can also do readthedocs.

callummole:
Closes #166

PR #436, opened 2023-02-02
Implement stories search bar

helendduncan:
Implementation of a search function to return (case insensitive) matches to the search from moderated stories title, experience, or suggestion

Uses GET (instead of POST)

Filters triggering stories if checkbox ticked

### DIFF
--- a/server/apps/main/templates/main/experiences_page.html
+++ b/server/apps/main/templates/main/experiences_page.html
@@ -42,11 +42,17 @@
 
 <section id="search-section">
   <div class="container search">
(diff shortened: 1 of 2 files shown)


callummole:
@helendduncan good work. I think that this should a `GET` request, rather than a `POST`, because we are retrieving data rather than changing the database. For example, when you type into google it adds the search to the url. See also the tail end of [here](https://learndjango.com/tutorials/django-search-tutorial) for how to do this in django

callummole (approved):

PR #437, opened 2023-02-03
Reformat datetime for My Stories

helendduncan:
Currently the creation date is read from the OH files as a string and is in an unnecessarily long form 

> 2023-01-30T15:40:39.857995Z

This inelegant solution reformats the string to a more compact (and readable) form

> 30-Jan-2023 15:40

A nicer solution would be to edit the time format in `settings.py` but I am not sure what else uses it and it may cause problems elsewhere in the code - what do you think @callummole?

### DIFF
--- a/server/apps/main/views.py
+++ b/server/apps/main/views.py
@@ -434,10 +434,20 @@ def get_review_status(files):
 
     return statuses
 
+def reformat_date_string(context):
+    """
+    Reformat the sting that contains the datetime to a more usable format
+    """
+    old_format = '%Y-%m-%dT%H:%M:%S.%fZ'
+    new_format = '%d-%b-%Y %H:%M'
+    for file in context['files']:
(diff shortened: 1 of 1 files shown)


gedankenstuecke:
Sorry for jumping in here, but maybe a "nicer" solution is to [use the _humanize_ functions](https://docs.djangoproject.com/en/4.1/ref/contrib/humanize/#naturaltime), which convert the timestamps into _X minutes/hours/days ago_. 

This solution would also help to bypass all issues around time zones etc. I'm 99% sure the time stamps OH are always in UTC, so we would have to do some magic on the frontend to convert it to user's local time otherwise. I guess for half the year there would be no issues in the UK, but once we are on DST people might get confused.

callummole:
Closing this as I believe this has been addressed in #438.

Issue #445, opened 2023-02-10
Consistent behaviour for moderation for Public->Private->Public stories


helendduncan:
### Summary

As noted by @gedankenstuecke. When flipping an existing experience from public to private and back, the moderation page crashes in a way that was observed when there were PublicExperience objects that did not have a UUID.

### Deadline 
Feb 2023
### What needs to be done?

- [ ] Investigate the cause
- [ ] Modify code to fix
- [ ] Additional: Think of tests that we can apply to check for expected behaviour

### Who can help?
@helendduncan 

---

gedankenstuecke:
Okay, I think I found out what causes the issue: 

When creating a new experience we call the `update_public_experience_db` function like this in the `share_experience` view: https://github.com/alan-turing-institute/AutSPACEs/blob/267a7aef5e34b96c6f1d08e30444c66132b8df9a/server/apps/main/views.py#L76

As part of this we set the `uuid=uuid` in the function call. The `update_public_experience_db` has a parameter, but then the actual creation of the public experience is the following: 

https://github.com/alan-turing-institute/AutSPACEs/blob/267a7aef5e34b96c6f1d08e30444c66132b8df9a/server/apps/main/views.py#L129

We don't explicitly set the different model parameters from the `PublicExperience` model, but rather use the `data` dict which now includes the `uuid`. The problem is that the `PublicExperience` model still expects the key to be called `experience_id` (see below). 

https://github.com/alan-turing-institute/AutSPACEs/blob/267a7aef5e34b96c6f1d08e30444c66132b8df9a/server/apps/main/models.py#L10

As a result the `experience_id` field always remains empty. This breaks the moderation form, but also leads to the problem that we currently can only store a single public experience as the `experience_id` is the primary key. 😂 

**Potential solutions**: 
1. rename the `experience_id` in the model to `uuid`, then the current calls should work 
2. explicitly rename the `uuid` to `experience_id` in the function call should also work

gedankenstuecke:
Closed by #446

PR #446, opened 2023-02-13
Store UUID in public experiences correctly

gedankenstuecke:
This closes  #445 

I noticed that it was only the passing of the `uuid` to the creation method that was missing to allow the moderation to work as expected. The second change was to then remove the potentially duplicate UUID that would be present when editing existing experiences.

### DIFF
--- a/server/apps/main/templates/main/share_experiences.html
+++ b/server/apps/main/templates/main/share_experiences.html
@@ -210,18 +210,20 @@
 
     {% if readonly %}
       </form>
-      <div class="form-group">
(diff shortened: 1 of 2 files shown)


helendduncan:
This solution fixes the problem outlined in issue #445 

Two possible long-term solutions were proposed ([link to comment](https://github.com/alan-turing-institute/AutSPACEs/issues/445#issuecomment-1427855344)) I think that updating function call is the tidiest

helendduncan (approved):

PR #448, opened 2023-02-20
Save Experience History

### LINKED ISSUE #440: Logging activity on files.
### Summary

We do not track activity across a file's history. For auditing we should track this information.

### Deadline 

Feb 2023

### What needs to be done?

This is a suggested list. Please update as the task gets clearer.

- [x] Include an originally posted date (or originally/most recently made public)
- [x] Decide what actions should be logged
- [ ] Write a module that, given an action, returns a logging string.
- [x] Decide how to track the logs within a file. The simplest method would be to have a field `log` that is just a list of strings. But there might be more sophisticated or well-thought through ways out there.
- [ ] Decide how we want to display the logging information. For example, we could have a 'last edited' column in `my stories`

### Who can help?

@helendduncan 
@callummole

helendduncan:
Closes #440 
An `ExperienceHistory` model has been made which has a many-to-one relationship with the `PublicExperience`
The `ExperienceHistory` saves:
The type of change (Made public, edited, moderated)
When the change was made
Who made the change
Accompanying change text

A moderate experience form has been produced which includes space for moderators comments which will be logged in by the `ExperienceHistory`

This should work with the current framework but the logic and management of the code has not been streamlined yet.

Changes that will be enacted in upcoming PRs.

- [ ] Change the type of `changed_by` from string to OHMember
- [ ] Remove hardcoded arguments in update_public_experience_db function
- [ ] Use the moderation form to fix the tags issue highlighted in discussion #431


server/apps/main/forms.py
```diff
@@ -56,6 +56,12 @@ class ShareExperienceForm(forms.Form):
     moderation_status = forms.ChoiceField(choices = statuses, widget = forms.Select(), required=False)
     moderation_status.group = "hidden"
 
+    moderation_comments = forms.CharField(label='Moderator comments', max_length=500, strip=True, required=False,
```
gedankenstuecke, line 59: Do we actually need the moderation_comments in this form?

server/apps/main/forms.py
```diff
@@ -78,3 +84,89 @@ def clean_moderation_status(self):
             mod_status = 'not reviewed'
 
         return mod_status
+
+
+class ModerateExperienceForm(forms.Form):
+        
+    experience_text = forms.CharField(label='Please share your experience', max_length=500, strip=True,
```
gedankenstuecke, line 91: Similarly to including the `moderation_comments` above, I guess we don't actually need the `experience_text`, `difference_text` and `title_text` in this form, if we don't plan on allowing moderators to edit those.

server/apps/main/forms.py
```diff
@@ -78,3 +84,89 @@ def clean_moderation_status(self):
...
+                            widget=forms.TextInput(), required=False)
+    other.group = 2
+
+    # sharing options
+    viewable = forms.BooleanField(label = "Share on AutSPACE website", required=False)
+    viewable.group = 3
+    research = forms.BooleanField(label = "Share for research", required=False)
+    research.group = 3
+
+    # hidden field for moderation status
```
gedankenstuecke, line 135: These shouldn't actually be hidden in the moderation form, right?

gedankenstuecke (approved):
I left some comments for things that could be streamlined but otherwise works as expected! 👍

server/apps/main/forms.py
helendduncan: No we don't thought I had commented that out!

gedankenstuecke: Ah, and I see that currently we don't actually use the moderation form! `model_to_form` still uses the `ShareExperienceForm`, which explains why it actually showed up! :D

server/apps/main/forms.py
helendduncan: They don't need to be. At the moment template checks for hidden groups and displays them under certain conditions (moderation) which at the moment is useful as the moderation status is present in both the Original and ModerateExperience forms. Probably by the end of things this will be changed

gedankenstuecke: Ah, okay!

server/apps/main/forms.py
helendduncan: Yeah I've just changed that

server/apps/main/forms.py
helendduncan: No but I think it is useful to have them in "form-form" at the moment?

gedankenstuecke: Yep, I think for now that's okay. It was mostly for what we want to work on going forward. Happy to have that merged in as is for now :)

Issue #447, opened 2023-02-17
Trigger warnings not removed during moderation


gedankenstuecke:
**Describe the bug**
@callummole and @badgermind encountered this issue during the last community call and (see discussion #431). I've done some digging into it: 

Moderators are given the option to add & remove trigger labels for stories during the moderation step, but the removal does not propagate.

**To Reproduce**
1. Submit an experience with at least one trigger label selected
2. As a moderator go to the moderation view, remove one of the set labels and submit as moderted
3. The now published experience will still have the original trigger label set despite the moderator removing it
4. This can also be seen as a user in the "my stories" tab

**Expected behavior**
When a moderator removes a label, it should be properly saved.

## My investigation so far 

I've reproduced the problem locally in the current `main` branch. Interestingly, _adding_ new trigger labels by moderators works as expected, it is just the removal that does not work. 

It seems that the issue is that changing a checkbox from formerly checked to now un-checked does not properly propagate. 

I tried initializing the form with the existing pre-moderation model like such: 

```
moderated_form = ShareExperienceForm(request.POST, initial=model.__dict__)
```

When doing this, the output of `print(moderated_form.data)` leads to the following, for a model that pre-moderation had _"drugs"_ & _"negbody"_ checked and after moderation should have _abuse_ & _negbody_ but **not** _drugs_: 

```
 <QueryDict: {'csrfmiddlewaretoken': ['g2X7RZoWp0uki1yMHWhOjgjJ7MqSRyk36DOM52JlL7r5A9vJuTfQZHZy5jA8nqha'], 'abuse': ['on'], 'negbody': ['on'], 'other': [''], 'moderation_status': ['in review']}>
```

As we can see the _drugs_ label is absent here. But if we run `print(moderated_form.changed_data)` the output is: 

```
['experience_text', 'difference_text', 'title_text', 'abuse', 'drug']
```

Here it seems to identify that the labels for _drug_ and _abuse_ have changed. But then the `process_enabled_form_keys` function does not take into account that a label might have switched from `True` to `False`. 

We'll need to investigate how to take those changes into account as well.

PR #449, opened 2023-02-20
Ensure changes to trigger warnings made during moderation persist

helendduncan:
Closes issue #447 

Replaced `process_enabled_form_keys` with `process_trigger_warnings`.
Uses the newly produced `ModerateExperience` form. 
Form is sent to the function and cleaned, list of trigger warnings is given to it and any that are absent are filled as false.
Story details extracted in separate function and merged into a data dictionary that is sent to `update_public_experience_db`


gedankenstuecke (approved):
that's great! thanks for the quick fix @helendduncan

Issue #444, opened 2023-02-07
Expired tokens during moderation


gedankenstuecke:
**Describe the bug**
Depending on the time between a _User A_ submitting their experience and the moderation done by _User B_ happening, it can occur that the _Open Humans_ token for _User A_ is expired or that _User A_ has deauthorized Open Humans in the meanwhile. 

If that happens then a) the upload of the moderation status to Open Humans fails with a _token expired_ error and b) we currently don't handle that error. 

We should consider what to do in that case and at minimum implement a graceful failure.

gedankenstuecke:
After [looking over the `django-open-humans` code](https://github.com/OpenHumans/django-open-humans/blob/master/openhumans/models.py) I think that at least the _expired tokens_ issue should not occur, as the library automatically refreshes the tokens as appropriate. 

This leaves the issue of users having de-authorized or left AutSPACEs between submitting their experience and the subsequent moderation.

Issue #421, opened 2023-01-25
Tidy codebase


callummole:
### Summary

Catch-all issue for some code refactoring and tidying up. 

### Deadline 

Feb 2023

### What needs to be done?

Removing redundant code:
- [x] Unused url patterns 
- [x] unused views

- [x] Separate helper functions from views
- [x] remove unused html
- [x] abstract repeated html.
- [ ] check the app logic is in model, not views, to avoid repetition (https://django-best-practices.readthedocs.io/en/latest/applications.html#make-em-fat)

Coding style
- [x] Ensure readability and django coding style: https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/.
- [x] Add linter.
- [x] add comments.

### Who can help?

* @helendduncan 
* @callummole

callummole:
Adding this aspirationally to Milestone 2. Thought process is that once we have an MVP we will want to look more widely for contribution, user testing etc. To help with this the codebase should be clear and easy to follow.

gedankenstuecke:
I think we're now basically done with this one after @helendduncan already added the linter in the GH action setup!

PR #450, opened 2023-02-20
Make separate moderation view

gedankenstuecke:
The moderation view now is a custom template instead of re-using the existing write/edit template. Beyond for allowing for a more customized view generally, I also added a display of the edit history that @helendduncan wrote in #448!

### DIFF
--- a/server/apps/main/forms.py
+++ b/server/apps/main/forms.py
@@ -82,27 +82,7 @@ class ShareExperienceForm(forms.Form):
 
 class ModerateExperienceForm(forms.Form):
         
-    experience_text = forms.CharField(label='Please share your experience', max_length=500, strip=True,
-                                 widget=forms.Textarea(attrs={'placeholder':'Write your experience here, you can take as much or little space as you need.',
-                                                              'rows':'4',
-                                                              'class':'form-control'}))
-    experience_text.group = 1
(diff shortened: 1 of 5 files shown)


gedankenstuecke:
Looks like we successfully avoided merge conflicts @helendduncan! 🎉

gedankenstuecke:
I've added a little exception handling for the moderation to catch #444 and also changed the redirect of the moderation `POST` so that moderating an experience takes the moderator back to the moderation overview.

I think with that most of moderation issues are handled for now? :)

gedankenstuecke:
Also fixed: 
- buttons being 'doubled' in moderation list view due to use of nested `button` and `a href` 
- `edit` button in the `view experience` page being outside main `div`

helendduncan (approved):
In this PR the moderation form is updated to remove the story parameters. A moderate experiences page has been made (and the view updated) which displays the story as an integrated part of the page, vs a form as was before and includes the a log of the experience history. Exceptions were introduced that will allow a deletion of public experience objects if the user has deauthorised AutSPACEs but not deleted their stories locally. Post moderation, the moderator is now returned to the moderation page, not the story submitted page.

gedankenstuecke:
And merged in just in time for the Tidying (#421)!

PR #454, opened 2023-02-27
Move helper functions out of views.py

helendduncan:
Currently supporting functions and the standard view functions are both held in `views.py`.
This PR will
- [x] Ensure only functions with a return or redirect in `views.py`
- [x] Streamline and document the helper functions
- [x] Streamline and document the view functions
- [x] Remove references to unused urls

This is two of the checkboxes mentioned in #421 

The files edited in this PR have been formatted by [black](https://black.readthedocs.io/en/stable/)


server/apps/main/views.py
```diff
@@ -1,61 +1,120 @@
...
+def logout_user(request):
+    """
+    Logout user
+    """
+    if request.user.is_authenticated:
+        logout(request)
+    return redirect("index")
+
+
 def index(request):
```
gedankenstuecke, line 84: I'm actually not 100% sure we need both the `index` and the `overview`, do we? Both seem to be rendering the same `main/home.html` template (and since one of the last PRs we always ahve the `auth_url` available in all templates as I fixed that bug as well). 

So maybe for now we could simplify to just use one of them?

server/apps/main/views.py
```diff
@@ -1,61 +1,120 @@
...
+
+def navigation(request):
+    return render(request, "main/navigation.html")
+
+
+def what_autism_is(request):
+    return render(request, "main/what_autism_is.html")
+
+
+def footer(request):
```
gedankenstuecke, line 53: I think this could be removed, the templating takes care of properly loading the footer/navigation etc.

server/apps/main/views.py
```diff
@@ -1,61 +1,120 @@
...
+    Confirmation Page For App.
+    """
+    return render(request, "main/confirmation_page.html")
+
+
+def about_us(request):
+    return render(request, "main/about_us.html")
+
+
+def navigation(request):
```
gedankenstuecke, line 45: I think this could be removed, the templating takes care of properly loading the footer/navigation etc.

server/apps/main/views.py
```diff
@@ -1,61 +1,120 @@
...
+
+def what_autism_is(request):
+    return render(request, "main/what_autism_is.html")
+
+
+def footer(request):
+    return render(request, "main/footer.html")
+
+
+def signup_frame4_test(request):
```
gedankenstuecke, line 57: this seems unused and could be removed I think

server/apps/main/urls.py
```diff
@@ -1,47 +1,34 @@
...
-    path('confirm_page/', views.confirmation_page, name="confirm_page"),
-    path('about_us/', views.about_us, name="about_us"),
-    path('what_autism_is/', views.what_autism_is, name="what_autism_is"),
-    path('navigation/', views.navigation, name="navigation"),
-    path('footer/', views.footer, name="footer"),
-    path('registration/', views.registration, name="registration"),
-
+    re_path(r"^$", views.index, name="index"),
+    re_path(r"^signup/?$", views.signup, name="signup"),
+    re_path(r"^signup1/?$", views.signup_frame4_test, name="signup_frame4_test"),
```
gedankenstuecke, line 9: I think this is unused effectively

server/apps/main/urls.py
```diff
@@ -1,47 +1,34 @@
...
+    path("delete/<uuid>/<title>/", views.delete_experience, name="delete_exp"),
+    path("share_exp/", views.share_experience, name="share_exp"),
+    path("edit/<uuid>/", views.share_experience, name="edit_exp"),
+    path("moderate/<uuid>/", views.moderate_experience, name="moderate_exp"),
+    path("view/<uuid>/", views.view_experience, name="view_exp"),
+    path("my_stories/", views.my_stories, name="my_stories"),
+    path("confirm_page/", views.confirmation_page, name="confirm_page"),
+    path("about_us/", views.about_us, name="about_us"),
+    path("what_autism_is/", views.what_autism_is, name="what_autism_is"),
+    path("navigation/", views.navigation, name="navigation"),
```
gedankenstuecke, line 31: navigation and footer can go as well :)

gedankenstuecke (approved):
I had a look over it and also tested the website, everything still works and the split into views and helpers is a lot more readable, really well done @helendduncan! 

I flagged some parts which could be removed as they aren't used, but we can also do that in a later PR if you prefer to keep this one simpler!

server/apps/main/views.py
helendduncan: Removed

server/apps/main/urls.py
helendduncan: Removed

server/apps/main/urls.py
helendduncan: Removed

server/apps/main/views.py
helendduncan: Removed

PR #467, opened 2023-03-23
add more helper tests

gedankenstuecke:


server/apps/main/tests/test_helpers.py
```diff
@@ -54,4 +55,44 @@ def test_moderator_check(self):
...
+        # Create mock set of tags 
+        # Mock tags include 3 true items from the list, 2 false items from the list
+        tag_data = {'abuse': True, 'violence': True,
+                    'drug': False, 'mentalhealth': False,'negbody': True,
+                    'other': '',
+                    'moderation_status': 'approved',
+                    'moderation_comments': ''
+                    }
+        returned_data = make_tags(tag_data)
+        print(returned_data)
```
gedankenstuecke, line 95: After writing this test, I think we actually might want to modify the `make_tags` function. Currently the call as done here returns the following: 

```
['abuse', 'violence', '', '', 'negative body', None]
```

The `None` is because the `moderation_status` response is not reflected in the key at all, the empty strings are due to how `make_tags` returns empty strings. I think more optimally it should only return the 3 actually present tags and nothing else? 

@helendduncan What do you think?

helendduncan: I don't think this function is used now - looking at it

gedankenstuecke: I think it's still used in the `upload` helper function which is called in the share experience-view (see https://github.com/alan-turing-institute/AutSPACEs/blob/c1796a78309f238f7ea963135032f00dbcba8140/server/apps/main/views.py#L122)

server/apps/main/tests/test_helpers.py
```diff
@@ -176,4 +180,23 @@ def test_delete_pe(self):
...
+            experience_text = 'EXP_TEXT', 
+            difference_text = 'DIFF_TEXT', 
+            title_text = 'TITLE',
+            experience_id = 'TEST_ID',
+            open_humans_member = self.moderator_user,
+            abuse = True, 
+            violence = True,
+        )
+        self.public_experience.save()
+        self.form = model_to_form(self.public_experience)
```
gedankenstuecke, line 200: I found another legacy bit that could be removed I think: The `disable_moderator` parameter and logic in the `model_to_form` for the `ModerateExperienceForm` seem to be superfluous now, since I modified the form to not include the actual experience text etc?

helendduncan: That sounds sensible - but let me wait until this evening to get it completely wrong again :-)

server/apps/main/tests/test_helpers.py
helendduncan: So in response to the original comment:

helendduncan: The use for the tags was to easily access all the information for the "my stories" page.The moderation status tag is used on a "per-story" basis still so given our current set up it would be good to have that stored in the metadata.

helendduncan: > I think to me in the example above we would like to switch the list of triggers to be `['abuse', 'violence', 'negative body']`, so I think this would be _keep only the trigger warning tags_, as those trigger labels that don't apply won't show up. Does that make sense? :)

Typing at the same time! Yes that does make sense and I agree. I think we do still need to preserve the other information in the metadata but they needn't be together.

gedankenstuecke: Yep, perfect! 

I guess the moderation status is stored in the `PublicExperience` object in our AutSPACEs database and additionally it's also explicitly stored in the top-level metadata object as `"moderation_status":"not reviewed"` and similarly the trigger labels are also stored in the top-level item. It just seems that we also copy much of that information into the `tags` list. :)

gedankenstuecke: I've made that change with the last commit, it now only returns the non-empty tags and the test accounts for that.

server/apps/main/tests/test_helpers.py
```diff
@@ -54,4 +62,215 @@ def test_moderator_check(self):
...
+        update_public_experience_db(self.pe_data, "foobar_id" ,self.moderator_user, 
+                                    change_type='Moderate',change_comments='I left a comment')
+        # check updated PE & latest PEH object
+        pe = PublicExperience.objects.get(experience_id='foobar_id')
+        self.assertEqual(pe.moderation_status, 'approved') # is now approved
+        self.assertEqual(len(ExperienceHistory.objects.filter(experience=pe)),3) # got 3 history entries now
+        peh = ExperienceHistory.objects.all().order_by('changed_at')[2] 
+        # check details of PEH & PE
+        self.assertEqual(pe.open_humans_member.oh_id,self.non_moderator_user.oh_id) # exp owned by owner
+        self.assertEqual(peh.changed_by.oh_id,self.moderator_user.oh_id) # exp last changed by moderator
```
gedankenstuecke, line 274: This test currently fails, but I think this is actually a mistake in how the function is written: 

The `update_public_experience_db` function currently always expects to be fed the _owner_ of the experience as the `ohmember` parameter. This is even true when a moderator changes the flag from `not reviewed` to `approved`. 

This is problematic, as the `update_public_experience_db` function uses the `ohmember` parameter to determine who made an action: https://github.com/alan-turing-institute/AutSPACEs/blob/bf004e2af946c663dff7226fc672aadcad94a362/server/apps/main/helpers.py#L261

As a consequence, any moderation action is currently stored as if it had been performed by the writer of the action. I think the easiest way to fix this would be to pass over the `request` parameter from the views into the helper function to fetch the user from there.

gedankenstuecke: I've fixed this in commit #d06fcd5

gedankenstuecke:
Jointly this should now basically provide a nearly complete code coverage for all the helper functions in `helpers.py`. It also addresses some minor issues found with the existing functions that were uncovered as part of writing the tests!

helendduncan:
Good catch on the moderation - I think it makes sense but if you are happy to wait until the weekend for a review I can have a proper look (and steal ideas)?

gedankenstuecke:
Most definitely, please do!

PR #469, opened 2023-03-28
Poetry update

llewelld:
This change incorporates Helen's poetry update changes and adds some additional comments for the benefit of users running with M1 macs.

I've tested this on an M1 mac and on an x64 Linux machine and it worked fine on both, but I've not been able to test it on Windows.

### DIFF
--- a/config/.env.template
+++ b/config/.env.template
@@ -17,6 +17,9 @@ TLS_EMAIL=webmaster@test.com
 # Generate yours with:
 # python3 -c 'from django.utils.crypto import get_random_string; print(get_random_string(50))'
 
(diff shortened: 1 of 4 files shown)


gedankenstuecke (approved):
That's some amazing work! @helendduncan and I have been struggling a bit getting it to run at different points in time. I just checked it out and this works perfectly, and it's great that it's now directly part of the docker script!

llewelld:
Thanks for reviewing and merging it. I want to emphasize though that @helendduncan deserves the credit: I just rolled it up into a PR. I'm very happy I can make use of it though.

PR #470, opened 2023-03-29
Test moderation views

gedankenstuecke:
- [x] check non-logged in users _can't_ access moderation views
- [x] check non-moderation users _can't_ access moderation views
- [x] check moderators _can_ access moderation views

### DIFF
--- a/server/apps/main/tests/fixtures/moderate_experience.yaml
+++ b/server/apps/main/tests/fixtures/moderate_experience.yaml
@@ -1,0 +1,275 @@
+interactions:
+- request:
+    body: project_member_id=12345678&file_basename=test-test-test.json
+    headers:
+      Accept:
+      - '*/*'
+      Accept-Encoding:
+      - gzip, deflate
+      Connection:
+      - keep-alive
+      Content-Length:
+      - '82'
+      Content-Type:
+      - application/x-www-form-urlencoded
+      User-Agent:
+      - python-requests/2.25.1
+    method: POST
+    uri: https://www.openhumans.org/api/direct-sharing/project/files/delete/
+  response:
+    body:
+      string: '{"ids":[63683438]}'
+    headers:
+      Allow:
+      - POST, OPTIONS
+      Cache-Control:
+      - max-age=0, no-cache, no-store, must-revalidate
+      Connection:
+      - keep-alive
+      Content-Length:
+      - '18'
+      Content-Type:
+      - application/json
+      Date:
+      - Wed, 29 Mar 2023 16:44:27 GMT
+      Expires:
+      - Wed, 29 Mar 2023 16:44:27 GMT
+      Server:
+      - gunicorn/20.0.4
+      Vary:
+      - Accept, Authorization, Cookie, Origin
+      Via:
+      - 1.1 vegur
+      X-Frame-Options:
+      - SAMEORIGIN
+    status:
+      code: 200
+      message: OK
+- request:
+    body: null
+    headers:
+      Accept:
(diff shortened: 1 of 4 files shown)


server/settings/environments/development.py
```diff
@@ -15,6 +15,9 @@
 
 DEBUG = True
 
+# Disable axis in development (which is also used for testing!)
+AXES_ENABLED = False
```
gedankenstuecke, line 19: `AXES` is for checking suspicious logins, which only applies to production. In development/tests needs to be disabled in order to allow test runners to work.

server/apps/main/views.py
```diff
@@ -353,4 +353,4 @@ def moderate_experience(request, uuid):
                 },
             )
     else:
-        redirect("index")
+        return redirect("index")
```
gedankenstuecke, line 356: Fixed to now properly redirect to index for non-logged in and non-moderator users (rather than leading an error 500)

gedankenstuecke:
I would consider this now ready for review! I think it covers virtually all possibilities for moderation actions. 

In particular it tests some disruptive behaviour that should be blocked:
1. Guests/non-moderators trying to edit the moderation flags for any post
2. Moderators should be able to set moderation flags & trigger labels, but __not__ edit the experiences themselves. 

It also fixes some minor bug in handling exceptions. 

With this our test coverage doesn't look too bad: 

Overall: 

```
Name                                                                   Stmts   Miss  Cover
TOTAL                                                                    722     84    88%
```

For the main relevant files:

```
Name                                                                   Stmts   Miss  Cover
server/apps/main/forms.py                                                 71     12    83%
server/apps/main/helpers.py                                               91      3    97%
server/apps/main/models.py                                                27      2    93%
server/apps/main/templatetags/custom_tags.py                              12      3    75%
server/apps/main/urls.py                                                   4      0   100%
server/apps/main/views.py                                                124     55    56%
server/urls.py                                                            11      0   100%
```

server/apps/main/tests/fixtures/moderate_experience.yaml
helendduncan, line 1: Was this a VCR fixture or is this something else? Just curious

server/apps/main/tests/test_moderate_views.py
```diff
@@ -0,0 +1,219 @@
...
+    # test moderation pages as logged-out user
+
+    def test_view_moderate_list_logged_out(self):
+        """
+        Test moderate overview page is not accessible by logged-out users
+        """
+        c = Client()
+        response = c.get("/main/moderate_public_experiences", follow=True)
+        self.assertRedirects(response, "/", 
+                             status_code=302,target_status_code=200)
```
helendduncan, line 61: So here the redirect pathway is... moderate -> overview (as not authenticated and a moderator) -> index (as not authenticated)

helendduncan (approved):

server/apps/main/tests/test_moderate_views.py
gedankenstuecke: Yes, so for users who are not moderators or not logged in the view goes to `main:overview` (https://github.com/alan-turing-institute/AutSPACEs/blob/f2e62dc402cf5a22ca1994889e486108aa6455d9/server/apps/main/views.py#L273). 

But the `main:overview` view in teurn reirects to index for non-authenticated users (https://github.com/alan-turing-institute/AutSPACEs/blob/f2e62dc402cf5a22ca1994889e486108aa6455d9/server/apps/main/views.py#L95)

PR #474, opened 2023-04-04
Cleanup metadata stored on OpenHuman

### LINKED ISSUE #468: Clean up metadata stored on OH
### Summary

Currently a single experience stored in Open Humans contains the following metadata. 

```
{
      "id": 62938625,
      "basename": "296683b6-bc44-11ed-b1d0-0242ac130003.json",
      "created": "2023-03-06T17:27:26.395856Z",
      "datatypes": [],
      "download_url": "https://www.openhumans.org/data-management/datafile-download/62938625/?key=4d01af57-ec09-48bf-84a7-49648bbed2cc",
      "metadata": {
        "data": {
          "drug": false,
          "abuse": false,
          "other": "",
          "negbody": false,
          "research": false,
          "viewable": true,
          "violence": false,
          "title_text": "The second of the stories",
          "mentalhealth": true,
          "difference_text": "Mr. Bennet replied that he had not.",
          "experience_text": "“My dear Mr. Bennet,” said his lady to him one day, “have you heard that Netherfield Park is let at last?”",
          "moderation_status": "not reviewed"
        },
        "tags": [
          "",
          "",
          "",
          "mental health",
          "",
          "public",
          "non-research",
          null
        ],
        "uuid": "296683b6-bc44-11ed-b1d0-0242ac130003",
        "timestamp": "2023-03-06 17:27:25.680349",
        "description": "The second of the stories"
      },
      "source": "direct-sharing-1111",
      "source_project": "https://www.openhumans.org/api/public/project/1111/"
    },
```

While a good number of those items are from OH itself (`source`, `source_project`, `basename`, `id`, `download_url`, `'data_types`), the most rich metadata is created by AutSPACEs on upload. Currently this metadata is becoming a bit too redundant and large and we could strip it down: In particular the `experience_text` and `difference_text` can be quite large and we also store some of the metadata fields (e.g. the trigger labels both in `data` as well as the `tags`). 

### What needs to be done?

1. Streamlining the `data` dictionary within the metadata. Concretely, the `experience_text` and `difference_text` can be removed. The other data can/needs to stay as we use it for quickly populating the _my stories_ page. 

2. The `tags` list in the metadata dictionary can be cleaned up: It already won't return empty or `null` values anymore, but ultimately we only need two bits of information in it: a) is this experience marked as public/private and b) is it marked as for research or not?

### Who can help?

* Anyone feeling comfortable adjusting the functions in the code
* 

---

llewelld:
Fixes #468.

There was considerable overlap between the metadata and the file contents stored on OpenHuman. This change streamlines the metadata to remove the descriptive text. The descriptive text is now retrieved by downloading the file from OpenHuman instead.

### DIFF
--- a/server/apps/main/helpers.py
+++ b/server/apps/main/helpers.py
@@ -110,6 +110,7 @@ def get_review_status(files):
     return statuses
 
 
+# @vcr.use_cassette("tmp/get_oh_metadata.yaml", filter_query_parameters=['access_token'])
 def get_oh_metadata(ohmember, uuid):
     """Returns the metadata for a single file from OpenHumans, filtered by uuid.
 
@@ -133,6 +134,7 @@ def get_oh_metadata(ohmember, uuid):
 
     return file[0]
 
+# @vcr.use_cassette("tmp/get_oh_file.yaml", filter_query_parameters=['access_token'])
 def get_oh_file(url):
     """Returns a single file from OpenHumans.
 
@@ -150,6 +152,7 @@ def get_oh_file(url):
 
     return data
 
+# @vcr.use_cassette("tmp/get_oh_combined.yaml", filter_query_parameters=['access_token'])
 def get_oh_combined(ohmember, uuid):
     """Returns a dictionary combining file and metadata from OpenHumans, filtered by uuid.
 
--- a/server/apps/main/helpers.py
+++ b/server/apps/main/helpers.py
@@ -2,11 +2,12 @@ import datetime
 import json
 import io
 import uuid
+import requests
 from django.contrib.auth.models import Group
 from django.forms.models import model_to_dict
+from django.conf import settings
 from .forms import ModerateExperienceForm
 from .models import PublicExperience, ExperienceHistory
-
 
 def is_moderator(user):
     """Return membership of moderator group."""
@@ -109,8 +110,8 @@ def get_review_status(files):
     return statuses
 
 
-def get_oh_file(ohmember, uuid):
-    """Returns a single file from OpenHumans, filtered by uuid.
+def get_oh_metadata(ohmember, uuid):
+    """Returns the metadata for a single file from OpenHumans, filtered by uuid.
 
     Args:
         ohmember : request.user.openhumansmember
@@ -120,7 +121,7 @@ def get_oh_file(ohmember, uuid):
         Exception: If uuid belongs to more than one file.
 
     Returns:
-        file (dict): dictionary representation of OpenHumans file. Returns None if uuid is not matched.
+        metadata (dict): dictionary representation of OpenHumans file metadata. Returns None if uuid is not matched.
     """
     files = ohmember.list_files()
     file = [f for f in files if f["metadata"]["uuid"] == uuid]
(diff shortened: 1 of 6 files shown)


llewelld:
Some things to note about this PR:
1. This doesn't use any OAuth authentication when downloading the file from OpenHumans. I couldn't see anything in [the docs](https://django-open-humans.readthedocs.io/en/latest/modules/interactions.html) about this, but the [OpenHumans CLI implementation](https://github.com/OpenHumans/open-humans-api/blob/af3179576540589d5b4ca0675302d246e8006944/ohapi/utils_fs.py#L469) appears to work similarly. Is this correct?
2. When combinding the metadata and file data from OpenHumans, if there's a conflict between them (e.g. the same key is found in both), the items in the metadata will take priority. In theory they should always be the same anyway, but should this be reversed?

gedankenstuecke:
Thanks @llewelld, that's some great work! 

Regarding your questions: 

1. Indeed, the file download doesn't use the `access_token`. What happens instead behind the scenes is that the `exchange_member` call that is triggered by `member.list_files()` generates a signed download URL for each file. Those URLs are valid for 60 mins before expiring. 

2. Great question! Maybe we could include a quick check in the function that does the merging and trigger an exception if that case should ever happen? That way we can have a longer think about how to resolve it without now implementing potentially unexpected behaviour down the line!

server/apps/main/helpers.py
```diff
@@ -167,6 +202,21 @@ def make_tags(data):
     tags = [tag for tag in tags if bool(tag) == True]
     return tags
 
+def prepare_metadata(uuid, timestring, data):
+    # The following items will be stripped from the metadata dictionary
+    metadata_mask = [
+        "experience_text",
+        "difference_text",
+        "title_text"
```
gedankenstuecke, line 210: I think we should keep the title-text in the metadata, as the _my stories_ view at `/main/my_stories/` uses them for its table to display the title for all experiences of a user without having to download all files from OH.

server/apps/main/helpers.py
```diff
@@ -132,6 +133,46 @@ def get_oh_file(ohmember, uuid):
 
     return file[0]
 
+def get_oh_file(url):
```
gedankenstuecke, line 136: Would you be interesting in writing some tests for the new functions as well? I'm asking as I know that it can be a bit tricky as we need to have the "right" VCR cassettes to mock the external API responses! Also happy to help and walk you through it if that would be useful!

gedankenstuecke (changes requested):

llewelld:
Thanks for the comments @gedankenstuecke which I really appreciate. I've converted this to draft while I sort out the unit tests.

Thanks also for your explanations of 1 and 2. I'm glad 1 is the correct behaviour.

Concerning 2, triggering an exception would be fine for new stories (since it shouldn't then happen), however it will definitely then trigger if data that's previously been uploaded to OpenHumans is used after this is merged. Would that be an issue?

gedankenstuecke:
Oh, great point for #2, which I guess would be fine as we don't have any production data yet, but maybe annoying for development. So happy to have the merging happening as you described for overwriting the data!

server/apps/main/helpers.py
gedankenstuecke: Whops, you're absolutely right! The `title_text` of the metadata isn't used, instead it uses the `description`, which we set to be the `title_text` in the `upload` function! I think in that case it's okay to leave it as is, as the `description` field is mandatory metadata when uploading to OH. 

Though, maybe we should truncate the title text to the first 100 characters when coping it into the `description` field, as [according to our documentation on OH](https://www.openhumans.org/direct-sharing/oauth2-data-upload/), the field only accepts that many characters?

server/apps/main/helpers.py
llewelld: I've given it a go. I'm not sure if the VCR tests I've added are along the right lines, so would value your input.

gedankenstuecke: Yep, that looks all right, well done!

server/apps/main/helpers.py
gedankenstuecke: I have to admit I didn't know it either, I originally was about to recommend to just remove the `description` field but then thought that I should check the OH docs before doing so and that's when I saw it 😂

gedankenstuecke (approved):
That seems all fixed, well done @llewelld!

Issue #433, opened 2023-01-30
Add warning for moderator status reset


callummole:
### Summary

Related to #423. Currently when someone edits a previous experience there is no warning that this resets the moderation pipeline.

### What needs to be done?

- [x] add warning, to either the `Edit` button on `my_stories` and when viewing readonly experiences, _or_ during the form-editing page, _or_ both. The simplest way to do this will like be a JS pop-up.
- [ ] 

### Who can help?

* @callummole 
* @helendduncan 

---

PR #478, opened 2023-04-12
Add warning popup when editing resets the moderation pipeline.

mastoffel:
* Adds warning popup modal when viewable stories get edited again.   
* Tracks the "Share on AutSPACE website" tickbox, so that unchecking the box during editing prevents the popup from showing up (as there won't be a second moderation). 

Issue #433

### DIFF
--- a/server/apps/main/templates/main/share_experiences.html
+++ b/server/apps/main/templates/main/share_experiences.html
@@ -252,5 +252,5 @@
 
   <script src="{% static 'js/submit-warning.js' %}"></script>
 
-  {% endblock %}
+{% endblock %}
 
--- a/server/apps/main/templates/main/share_experiences.html
+++ b/server/apps/main/templates/main/share_experiences.html
@@ -218,13 +218,39 @@
     </div>
     {% else %}
       <div class='form-group'>
-        <input type="submit" class="btn btn-outline-dark btn-lg float-right" value="Submit">
+        <!-- Add an id and data attributes to the submit button -->
+          <button type="button" class="btn btn-outline-dark btn-lg float-right" id="submitBtn" 
+          data-uuid="{{ uuid }}" data-moderation-status="{{ moderation_status }}">Submit</button>
       </div>
     </form>
   </div>
     {% endif %}
 
 
-</section>
+  </section>
 
-{% endblock %}
+  <!-- Modal for the submit button -->
+  <div class="modal fade" id="submitModal" tabindex="-1" role="dialog" aria-labelledby="submitModalLabel" aria-hidden="true">
+    <div class="modal-dialog" role="document">
+      <div class="modal-content">
+        <div class="modal-header">
+          <h5 class="modal-title" id="submitModalLabel">Confirmation</h5>
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
+        </div>
+        <div class="modal-body">
+          Submitting your edited story will send it for moderation again. Do you want to proceed?
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-secondary" data-dismiss="modal">Cancel</button>
+          <button type="button" class="btn btn-primary" id="submitForm">Submit</button>
+        </div>
+      </div>
+    </div>
+  </div>
+
+  <script src="{% static 'js/submit-warning.js' %}"></script>
+
+  {% endblock %}
+
--- a/server/apps/main/views.py
+++ b/server/apps/main/views.py
@@ -151,15 +151,20 @@ def share_experience(request, uuid=False):
                 data = get_oh_combined(ohmember=request.user.openhumansmember, uuid=uuid)
                 form = ShareExperienceForm(data)
                 title = "Edit experience"
+                viewable = data["metadata"]["data"].get("viewable", False)
+                moderation_status = data["metadata"]["data"].get("moderation_status", "not reviewed")
 
             else:
                 form = ShareExperienceForm()
                 title = "Share experience"
+                viewable = False
+                moderation_status = "not reviewed"
 
             return render(
                 request,
                 "main/share_experiences.html",
-                {"form": form, "uuid": uuid, "title": title},
+                {"form": form, "uuid": uuid, "title": title,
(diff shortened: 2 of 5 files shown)


server/apps/main/templates/main/share_experiences.html
```diff
@@ -218,13 +218,38 @@ <h5>{{ subheader }}</h5>
...
+        </div>
+        <div class="modal-footer">
+          <button type="button" class="btn btn-secondary" data-dismiss="modal">Cancel</button>
+          <button type="button" class="btn btn-primary" id="submitForm">Submit</button>
+        </div>
+      </div>
+    </div>
+  </div>
+
+  <script src="{% static 'js/submit-warning.js' %}"></script>
```
llewelld, line 252: This is super-nice; I love the slick modal dialog.

static/js/submit-warning.js
```diff
@@ -0,0 +1,53 @@
+document.addEventListener('DOMContentLoaded', function () {
```
llewelld, line 1: Since jQuery is already being loaded for bootstrap, it would probably make sense to make use of it here too.

Using jQuery's [`.ready()`](https://api.jquery.com/ready/) is more robust than `DOMContentLoaded` (since it fires even if things are loaded in the wrong order).

static/js/submit-warning.js
```diff
@@ -0,0 +1,53 @@
...
+  let viewableCheckbox = document.getElementById('shareOnAutSPACEs');
+  
+  // Check that event listener is working
+  // if (viewableCheckbox) {
+  //   viewableCheckbox.addEventListener('change', function () {
+  //     console.log('viewableCheckbox checked:', viewableCheckbox.checked);
+  //   });
+  // }
+  
+  if (submitFormButton) {
```
llewelld, line 13: Using the jQuery [`$('#submitForm')`](https://api.jquery.com/jQuery/) instead of `getElementById` has the benefit that it never returns null, so using it would mean you could drop these checks.

static/js/submit-warning.js
```diff
@@ -0,0 +1,53 @@
...
+  let form = document.querySelector('.form-context');
+  let submitFormButton = document.getElementById('submitForm');
+  let viewableCheckbox = document.getElementById('shareOnAutSPACEs');
+  
+  // Check that event listener is working
+  // if (viewableCheckbox) {
+  //   viewableCheckbox.addEventListener('change', function () {
+  //     console.log('viewableCheckbox checked:', viewableCheckbox.checked);
+  //   });
+  // }
```
llewelld, line 11: Could the commented code be removed, or could it still be useful for testing later?

static/js/submit-warning.js
```diff
@@ -0,0 +1,53 @@
...
+    }
+    
+    // When the submit button is clicked in the modal popup, submit the form
+    submitFormButton.addEventListener('click', function () {
+      submitModal.hide();
+      form.submit();
+    });
+  }
+});
+
```
llewelld, line 46: This is really nice; works well. I would suggest switching to use jQuery, but the overall approach looks good to me.

server/apps/main/templates/main/share_experiences.html
```diff
@@ -218,13 +218,38 @@ <h5>{{ subheader }}</h5>
...
+
+  {% endblock %}
```
llewelld, line 254:
```suggestion
{% endblock %}
```
I think the extra indent was an accident here.

llewelld (changes requested):
This is really nice. I have just a few comments:

1. Could you rebase your changes on the latest changes in main please? After that there are a couple of further small manual changes you'll need to make to get things working again. I've added comments to the code about what needs changing.
2. The functionality is great; my suggestion would be to use jQuery in your JavaScript unless there's a reason not to that I missed. One thing I noticed in testing was that to get jQuery to work, you'll also have to move the `{% include 'main/partials/jquery.html' %}` line in `application.html` to the `head` section so it's loaded before your code.
3. I'm not sure what the project preference is, but my personal preference would be to squash your changes into a single commit so that the history is kept clean. If that's not the project's approach, then please feel free to ignore this!

server/apps/main/views.py
```diff
@@ -144,15 +144,20 @@ def share_experience(request, uuid=False):
                 data = get_oh_file(ohmember=request.user.openhumansmember, uuid=uuid)
                 form = ShareExperienceForm(data["metadata"]["data"])
                 title = "Edit experience"
+                viewable = data["metadata"]["data"].get("viewable", False)
+                moderation_status = data["metadata"]["data"].get("moderation_status", "not reviewed")
```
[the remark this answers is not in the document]
mastoffel: Done, thanks!

static/js/submit-warning.js
mastoffel: Great suggestion, I've rewritten with jQuery.

server/apps/main/templates/main/share_experiences.html
mastoffel: Changed.

static/js/submit-warning.js
mastoffel: Good to know. Changed to jQuery and removed the if's.

server/apps/main/templates/main/share_experiences.html
```diff
@@ -218,13 +218,38 @@ <h5>{{ subheader }}</h5>
     </div>
     {% else %}
       <div class='form-group'>
-        <input type="submit" class="btn btn-outline-dark btn-lg float-right" value="Submit">
+        <!-- Add an id and data attributes to the submit button -->
+          <button type="button" class="btn btn-outline-dark btn-lg float-right" id="submitBtn" data-uuid="{{ uuid }}" data-moderation-status="{{ moderation_status }}">Submit</button>
```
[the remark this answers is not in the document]
mastoffel: Done.

mastoffel:
Great, thanks. I'm just missing the testing but will try to figure this out.

llewelld (approved):
Thanks for making the changes, this LTGM now: a really nicely implemented addition.

As we discussed, it probably makes sense to move the tests to a separate task given there's nothing there for this view yet, and I'll leave the decision about how to squash the commits (or not) to you, so I'm approving this now.

PR #484, opened 2023-04-17
Restructure moderation pages

### LINKED ISSUE #463: Improve moderation page
### Summary

Currently the moderation page that moderators see contains an accordion view that splits experiences into different categories. This approach will not work once we have a lot of experiences to moderate. 

### Deadline 
Mid-april 2023

### What needs to be done?

To improve the moderation overview page there are a number of improvements

- [ ] Split the moderation view into at least 3 sub-views: unmoderated/approved/rejected
- [ ] Allow pagination for each of the 3 sub-categories
- [ ] Add search functionality 

### Who can help?

* @gedankenstuecke 
* @helendduncan 

---

llewelld:
Restructures the moderation pages:

1. Splits the moderation view into three pages: pending, approved and rejected
2. Adds pagination for each of the categories
3. Adds search functionality

Fixes #463.

### DIFF
--- a/server/apps/main/helpers.py
+++ b/server/apps/main/helpers.py
@@ -6,6 +6,9 @@ import requests
 from django.contrib.auth.models import Group
 from django.forms.models import model_to_dict
 from django.conf import settings
+from django.core.paginator import Paginator
+from django.db.models import Q
+from django.shortcuts import redirect, render
 from .forms import ModerateExperienceForm
 from .models import PublicExperience, ExperienceHistory
 
@@ -23,6 +26,7 @@ def is_moderator(user):
 def model_to_form(model, disable_moderator=False):
     """Converts Model to form."""
     model_dict = model_to_dict(model)
+    model_dict["moderation_prior"] = model_dict["moderation_status"]
 
     form = ModerateExperienceForm(
         {**model_dict, "viewable": True},  # we only moderate public experiences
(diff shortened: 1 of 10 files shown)


llewelld:
This doesn't remove the existing moderation page, since I thought it might be useful for comparison. If the changes are agreed, then I had planned to create a separate PR to remove it.

llewelld:
Rebased and made a small change to up the coverage.

helendduncan:
@llewelld - can I check where the modified pages are?

llewelld:
> I mean I can see the pages approved etc exist but is there a link in the nav bar or are we directly linking to them?

That's a good question. With the change here it will add a new drop down in the nav bar that will only appear if you have the moderator role, like this:

image

However, this approach is all open to discussion of course... the nav bar is arguably getting a bit crowded.

One other point I should have mentioned is that the current implementation has a "Google Form for Questions Collection" which I didn't include in this change. So there's an open question about where that should go too.

GeorgiaHCA:
I think this would be a great one to discuss tomorrow with James and Susanna @helendduncan?

helendduncan (approved):
I don't know that we ever discussed fully the search function in the moderation pages on Friday 21/04/23 meeting - but I think this looks good as is.
Query. In views.py we import the helper function `get_elided_page_range` but it isn't used (that I can tell) should this be removed?

llewelld:
> Query. In views.py we import the helper function `get_elided_page_range` but it isn't used (that I can tell) should this be removed?

You're absolutely right, thank you! I've removed it, it wasn't needed.

llewelld:
Rebased.

server/apps/main/helpers.py
```diff
@@ -364,3 +368,104 @@ def update_public_experience_db(data, uuid, ohmember, editing_user, **change_inf
 
     else:
         delete_PE(uuid, ohmember)
+
+def get_elided_page_range(current, total, on_each_side, on_ends):
```
gedankenstuecke, line 372: I guess we needed to add this ourselves as our Django version is so old that we don't have access to the now-build-in version of it? 🙈  

If that's the case, maybe we should generally try to see if we can upgrade to a more modern version of Django? If i read the poetry file correct we're currently 2 major releases behind?

gedankenstuecke (commented):
Beyond the one question regarding the Django version I think there's one big question of whether we should abstract out the moderation views to merge them into one: 

Currently we have 3 views and 3 templates for the `approved` `pending` and `rejected` states. But realistically they are all virtually identical I think? 

The main differences are 
- the queries we are running in the `views.py`
- the wording in the templates is slightly different to adapt to the state

My suggestion would be to have one view and one template and use that view in association with a URL parameter for deciding which filtering/words to fill in (e.g. by having the `urls.py` make sure that `/main/moderate/pending` and `/main/moderate/rejected` turn the `pending` and `rejected` into a parameter for the view – similar to how we are using the parameters for the `edit` view). 

That would allow us to have one view and adapt the query parameters inside the view, which is already done mostly in the `moderate_page` helper! Similarly, we can then have a single `moderate_experience.html` template which fills in the wording based on the context. Does that make sense? 

I'm agnostic on whether we should merge this PR as is though and then make those abstractions in a new PR or whether we go ahead in this one. Do you have a preference @llewelld?

server/apps/main/helpers.py
llewelld: Yes, exaxctly right. I've rebased these changes on top of your Django upgrade (many thanks for doing that), and as a result have now been able to switch the code to use Django's version of the method instead.

llewelld:
> My suggestion would be to have one view and one template and use that view in association with a URL parameter for deciding which filtering/words to fill in

Thanks for this suggestion. You're right, there was a lot of unecessary redundancy there. I've reworked the code based on the approach you described so that the same template is used for all three.

This is definitely better in my opinion. the only downside is if more divergent changes are needed between the three pages in future. If you don't envisage this, then I'd agree it's better to consolidate them.

In this case, I'd squash the two commits here before merging.

gedankenstuecke (approved):
That's great, thanks so much for making those changes @llewelld! I'd be happy to have this merged in at this point!

llewelld:
Thanks! I've squashed the commits and rebased, and will merge once the pipeline has passed.

Issue #488, opened 2023-04-21
Hide triggering stories by default


helendduncan:
### Summary

The default view of the (moderated, approved) experiences should be only non-triggering stories

~~No preview text for stories that contain content warnings~~

Users opt into viewing potentially triggering content. On a per-label basis (i.e only allow "abuse" trigger tag)
### Deadline 

May 2023

### What needs to be done?

- [x] Switch default behaviour in view stories
- [x] Investigate toggles/dropdowns/checkboxes
- [ ] Have options for user testing

### Who can help?

* @helendduncan 

---

### Updates

4th May 2023 - after meet up ([Google doc](https://docs.google.com/document/d/1mYcTs0kfh8HWK1X87Wc5whARrhH_Rm9xaw2a-uT6UFs/edit#heading=h.caqz9vddyyxh)):
Current implementation (checkboxes under search bar) is potentially confusing as you have to "search" to view triggering stories.
HD's other idea of a dropdown with multiple options seems to fall under the same sub-optimal implementation (searching). 
Suggested alternatives: toggles (dynamically update page)
#### Some selected relevant comments from 4th May meet up:
* Can it default to everything is clicked, and you can unclick what isn’t okay?
* May result in some people being triggered because they didn’t recognise the opt-out buttons
* Neither checkboxes nor drop-down may be what we want? Perhaps we need a toggle? - dynamically reload - this is an expected behaviour
* https://www.bootstraptoggle.com This is a random “bootstrap plugin” for the frontend that would provide such toggle-switches
* Be more explicit and declared - we have excluded triggering content/now you are opting in for triggering content
* Guidance for seeing people through: “We have excluded potentially triggering content, if you would like to see more content please choose which categories you would like to see”
* Add a text box about what is shown, and what happens when you select the triggers 
* Toggles vs. check boxes - when does the search refresh? Toggle could have a yes or no side on it to make it clear which selection opts in. Check boxes are a more ubiquitous UI 
* Consistency is the most important part so we need to be fully team check box or fully team toggle
* Worth user testing? Could be different for autistic users because more sensitivity to regularity and different

gedankenstuecke:
Closed by #492

PR #492, opened 2023-04-24
View stories: Default behaviour and explicit trigger warnings

helendduncan:
This PR addresses issue #488.

On loading the public experiences page, the default behaviour is now to hide all triggering content. Users actively opt into seeing triggering by either selecting the triggers they are happy to see, or by selecting all.

Page automatically updates when changes to the triggering options are made.

There is a sister-branch [iss488_story_view_toggle](https://github.com/alan-turing-institute/AutSPACEs/tree/iss488_story_views_toggle) which has the same functionality but toggles instead of checkboxes. Interested in feedback for user experience/intuition for this.

The methods in helpers.py are inelegant - the use of exclude should be replace the current filtering methods with a minor refactoring of `list_public_experiences`.


helendduncan:
Other option for this (with toggles is branch [*_toggle](https://github.com/alan-turing-institute/AutSPACEs/tree/iss488_story_views_toggle)

static/js/shared-stories-filtering.js
```diff
@@ -0,0 +1,27 @@
...
+
+function isAllChecked(){
+    if(all.checked){
+        pressSubmit();
+    }
+}
+
+function pressSubmit() {
+    searchForm.submit();
+}
```
llewelld, line 27: One nice way (I think) to simplify this would be to use jQuery's ability to act on all elements with the same class simultaneously. That way you could reduce your code to something like this:
```javascript
$(function() {
  $(".custom-control-input").on("change", pressSubmit)

  function pressSubmit() {
      $("#search-form").submit()
  }
})
```
This skips the `isAllChecked` test, because I think in practice that will never apply (because the `all` checkbox is always unchecked when the page is loaded).

server/apps/main/helpers.py
```diff
@@ -438,3 +439,83 @@ def choose_moderation_redirect(moderation_prior):
...
+        if trigger == "mentalhealth":
+            experiences =  experiences.filter(Q(mentalhealth=True) | Q(mentalhealth=False))
+        if trigger == "negbody":
+            experiences =  experiences.filter(Q(negbody=True) | Q(negbody=False))
+        if trigger == "other":
+            experiences = experiences.filter(~Q(other="") | Q(other=""))
+    
+    return experiences
+
+def no_show_filer(experiences, triggers_to_show):
```
llewelld, line 487:
```suggestion
def no_show_filter(experiences, triggers_to_show):
```
A typo, I think?

server/apps/main/helpers.py
```diff
@@ -438,3 +439,83 @@ def choose_moderation_redirect(moderation_prior):
...
+
+    # t_list = [PublicExperience._meta.get_field(trigger) for trigger in triggers_to_show]
+
+    return(triggers_to_show)
+
+def active_exlusion(experiences):
+    # Use the exclude method
+    pass
+
+def show_filiter(experiences, triggers_to_show):
```
llewelld, line 460:
```suggestion
def show_filter(experiences, triggers_to_show):
```
A typo, I think?

server/apps/main/helpers.py
```diff
@@ -438,3 +439,83 @@ def choose_moderation_redirect(moderation_prior):
...
+    triggers_to_show = list(all_triggers.intersection(allowed_triggers))
+    # invert this to get triggers to hide
+
+    # t_list = [PublicExperience._meta.get_field(trigger) for trigger in triggers_to_show]
+
+    return(triggers_to_show)
+
+def active_exlusion(experiences):
+    # Use the exclude method
+    pass
```
llewelld, line 458: It looks like `active_exlusion` is no longer used. Can it be safely removed?

server/apps/main/helpers.py
```diff
@@ -438,3 +439,83 @@ def choose_moderation_redirect(moderation_prior):
...
+    """
+    Expand the QuerySet. 
+    Recieves the filtered(allowed) stories and filters by the model fields which
+    correspond to the triggering content labels
+    """
+
+    experiences = show_filiter(experiences, triggers_to_show)
+    experiences = no_show_filer(experiences, triggers_to_show)
+
+    return experiences
```
llewelld, line 521: You mention in the description that this is an inelegant approach. I'm not sure what the alternative would be, but for what it's worth, these look all very clear and elegant to me.

llewelld (commented):
This is really nice, works well and is clear to me. I'm not sure whether this is the right place to mention these, but I wanted to make a few general points:

1. I tested out both the 'checkbox' and 'toggle' branches, and aesthetically I much prefer the toggles. I do find the alternative arguments (consistent, standardisation) compelling too, but personally I didn't find the toggles disorientating.
2. As you mentioned in your demo earlier, currently the "Allow all triggering content" checkbox disables itself after you click it. It would be really nice if it updated itself dynamically (i.e. became checked if and only if all of the other checkboxes are checked). That way I could use it to *disable* all of the checkboxes as well as *enabling* them all (e.g. click it once to enable them all, click it again to disable them all).
3. I think this is going to be a controversial point, but for me, the checkboxes intuitively feel the wrong way around. By which I mean I'd expect them to *hide* content when they're active, rather than *show* content when they're active. This is especially the case for stories that have mutliple tags assigned to them.

This is all very opinionated stuff, and separate from the technical implementation here, which I think is really nicely put together.

gedankenstuecke (approved):
That's great Helen and works so well! 

I've just pushed some minor edits to the declarations in the front-end, where some minor bootstrap issues occurred in particular for mobile views that didn't have enough spacing and where text ran out of the viewport. Sorry for not adding them to the review as suggestions, but as it required some tweaking that seemed easier! 

I have one minor comment/question for the "allow all" toggle but I'll make a new issue for that one so that @mastoffel doesn't block on this!

PR #493, opened 2023-04-28
Upgrade to Django 4 & Python 3.9

gedankenstuecke:
Following my review of PR #484 I thought it would maybe be good if we updated our dependencies to use a more modern version of both Python (partially to avoid the M1 issues some of us encountered) but in particular Django (as we seemed to be 2 major releases behind). 

I've done the minimum necessary changes to both the Python code itself as well as to the Dockerfile. One related change is the Postgres update, as we again used a very outdated one that modern Django didn't support.


gedankenstuecke:
After forgetting to update the workflow files, the tests/CI now also uses the correct python/postgres versions!

docker-compose.yml
```diff
@@ -7,10 +7,10 @@
 # Should be used together with `docker/docker-compose.prod.yml`
 # in production.
 
-version: "3.6"
+version: "3.6.1"
 services:
   db:
-    image: "postgres:9.6.9-alpine"
+    image: "postgres:15.2-alpine"
```
llewelld, line 13: Just a couple of things I noted in relation to the postgres upgrade:
1. The databases don't seem to be compatible between the two, so I had to delete mine to get it to work. Does there need to be a migration strategy? For me it wasn't a problem because there was nothing in the database I was concerned about keeping, so I just deleted it.
2. The latest postgres docker image doesn't seem to allow `POSTGRES_PASSWORD` to be left blank. Would it be worth adding a note to the `config/.env.template` file about this?

llewelld (commented):
Everything I tested was working after the upgrade. This is really nice and a very good idea; thank you.

docker-compose.yml
gedankenstuecke: 1. Indeed, one would need to update the storage that the Docker container creates to the newer postgres version. This is possible and would be necessary if we had production data, but I guess as we don't at this stage we're good! 😉 
2. Indeed, good point to leave a comment!

server/settings/components/common.py
```diff
@@ -115,7 +114,7 @@
 LANGUAGE_CODE = 'en-us'
 
 USE_I18N = True
-USE_L10N = True
+# USE_L10N = True
```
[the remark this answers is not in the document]
gedankenstuecke: Excellent point! Fixed that in the latest commit by removing it!

gedankenstuecke:
Thanks @llewelld! I've made those suggested changes in the last commit, so should hopefully be good to go and will save us a lot of pain going forward! :)

llewelld (approved):
This looks good to me, thank you!

gedankenstuecke:
Great, thanks!

PR #501, opened 2023-05-05
update/add contributors

gedankenstuecke:
It's been a while since we updated the list and I've added some of our new contributors and also updated the contributions for some folks where they had grown. 

I'm sure I've missed some additional contributions, so happy to take feedback on my oversights! :)


README.md
```diff
@@ -1,6 +1,6 @@
 # AutSPACEs
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
-[![All Contributors](https://img.shields.io/badge/all_contributors-36-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/badge/all_contributors-38-orange.svg?style=flat-square)](#contributors-)
```
llewelld, line 3:
```suggestion
[![All Contributors](https://img.shields.io/badge/all_contributors-38-orange.svg?style=flat-square)](#contributors)
```
It looks like this gained an accidental extra `-` at the end... at least this seems to break the forward link for me (but maybe it's there for a reason that I'm not appreciating?).

llewelld (approved):
Thanks for adding us! I've made just one very small comment, but this looks good to me.

README.md
gedankenstuecke: Great catch! I've fixed it! And it seems that the CLI tool for `all-contributors` adds this, which is weird. I'll check if that's a reported bug!

Issue #455, opened 2023-02-27
Add pagination to my_stories, shared_stories pages, and search results


helendduncan:
### Summary
When the db contains a significant number of stories, it will be unwieldy to have them all page. Suggest use of pagination to limit the number of stories visible per page
### Deadline 
March 2023
### What needs to be done?
Update appropriate templates to allow for pagination
Decide on a sensible number of stories per page
Update views.py
### Who can help?

* 
* 

---

gedankenstuecke:
Missing: Pagination on search/overview page

PR #507, opened 2023-05-16
Add pagination to mystories

mastoffel:
Paginates the different tables under the my-stories tabs. 

Any suggestions to improve this are very welcome. At the moment I made a separate pagination.html template to not repeat the pagination logic for every tab. 

Addresses #455


mastoffel:
Still needs tests.

server/apps/main/views.py
```diff
@@ -322,22 +326,43 @@ def moderation_list(request):
...
-        statuses = get_review_status(context["files"])
-        context = {**context, **statuses}
+
+        # Define the number of items per page
+        items_per_page = 10
+
+        # For each category, create a Paginator and get the page requested by the user
+        paginator_public = Paginator(filter_by_tag(filter_by_moderation_status(files, "approved"), "public"), items_per_page)
+        page_public = request.GET.get('page_public')
+        context['public_stories'] = paginator_public.get_page(page_public)
```
llewelld, line 346: This is really nice, but currently the pagination style doesn't quite match up with the pagination style on the moderation pages. In particular, it seems to be missing the first/last and ellipses entries. I.e. what I'd expect to see is something like this:

image

Whereas in practice I'm seeing this (no ellipses, no first entry):

image

It would be nice if these could be added in I think. It should be possible to use [`Paginator.get_elided_page_range()`](https://docs.djangoproject.com/en/4.2/ref/paginator/#django.core.paginator.Paginator.get_elided_page_range) for this.

llewelld (commented):
This is really nice; great stuff. I've added just a few small comments.

server/apps/main/views.py
```diff
@@ -322,22 +326,43 @@ def moderation_list(request):
...
     if request.user.is_authenticated:
-        context = {"files": request.user.openhumansmember.list_files()}
+        files = request.user.openhumansmember.list_files()
+        context = {"files": files}
         context = reformat_date_string(context)
-        statuses = get_review_status(context["files"])
-        context = {**context, **statuses}
+
+        # Define the number of items per page
+        items_per_page = 10
```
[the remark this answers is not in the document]
gedankenstuecke: Agree, I guess we'll have pagination on a number of pages (moderation, public experiences, "my stories"). For now it might be easiest to have them all use the same parameter. 

Going forward we might want to change the parameter to different ones for different pages, but even then it would probably be nice to have them set up in the settings!

static/js/store-tab.js
```diff
@@ -0,0 +1,18 @@
...
+    // get the last tab if it exists
+    var lastTab = localStorage.getItem('lastTab');
+    
+    if (lastTab) {
+      $('#' + lastTab).tab('show');
+    } else {
+      // Set the first tab to active if no active tab exists
+      $('a[data-toggle="tab"]:first').tab('show');
+    }
+  });
```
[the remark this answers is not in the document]
gedankenstuecke: absolutely, well done!

server/apps/main/views.py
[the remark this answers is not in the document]
mastoffel: Done, thanks!

static/js/store-tab.js
```diff
@@ -0,0 +1,18 @@
+// This file is used to store the last active tab in the browser's local storage
+$(function() {
+    // when a tab is clicked
+    $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
+      // save the latest tab using a cookie
+      localStorage.setItem('lastTab', $(this).attr('id'));
```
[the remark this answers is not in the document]
mastoffel: Good point, done!

server/apps/main/views.py
mastoffel: Yes, again good point and thanks for hint to` get_elided_page_range()`. Changed that.

mastoffel:
Thanks everyone! Now that pagination takes a few more lines of code, I put it in a function in helpers.py.

server/apps/main/tests/test_story_filters.py
```diff
@@ -52,25 +50,30 @@ def setUp(self):
...
+        filtered_files = filter_in_review(self.files)
+        self.assertEqual(len(filtered_files), 2)
+        
+
+class TestPaginateMyStories(TestCase):
+    def setUp(self):
+        self.factory = RequestFactory()
 
-        self.assertEqual(int(rendered_string), 2)
+    def test_paginate_my_stories(self):
```
gedankenstuecke, line 69: I don't think this currently doesn't actually test whether our pagination works, right? 

As the `Paginator` is being passed a `range(30)` the paginated items are just the numbers 0-29 I think, so the `request` from the factory is in a way superfluous.

gedankenstuecke (commented):
Looks great overall, thanks so much @mastoffel!  

The only comment I had was regarding the test! Would you like to work on implementing the testing as part of this PR or prefer later on? 🙂

server/apps/main/tests/test_story_filters.py
mastoffel: Ah yes, of course, thanks! I'll work on the test now as part of this PR!

gedankenstuecke: Great, thanks!

mastoffel: Ok! Added the (hopefully) proper pagination tests.

gedankenstuecke: Cool, will have a look now!

server/apps/main/tests/test_my_stories.py
```diff
@@ -0,0 +1,94 @@
...
+
+        
+# ------ test pagination -------- #
+
+class TestPaginateMyStories(TestCase):
+    def setUp(self):
+        self.factory = RequestFactory()
+
+    def test_paginate_my_stories(self):
+        request = self.factory.get('/my-stories/?page_test=2')  # mimic a request for the second page
```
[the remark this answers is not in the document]
mastoffel: Yes that's what I thought. Or do you think it's still necessary to have VCR and proper log in?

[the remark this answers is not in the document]
gedankenstuecke: I think it's fine like it is, I was just trying to understand why the request and wasn't sure I had 100% followed. But happy to have that merged otherwise I think 👍

PR #509, opened 2023-05-18
Add moderation replies using freetext field

llewelld:
Allows a reply to be included with each moderation which is visible in the history table on the moderation page, and to the experience author on the edit story page.

In this implementation, moderators can see all previous replies in the table, but the author can see only the most recent reply.

------

This is a slightly amended version of the previous PR #494 which allows for freetext entry rather than using the moderation menu. Differences from the previous version are:

1. The moderator reply and moderator comments are now full-width text fields to improve clarity.
2. The moderator reply is now shown on both the view story and edit story pages.


llewelld:
PR #494 now builds on this to switch the freetext field for a moderator menu. Both versions amend the database (so migration is needed), but both in the same way (i.e. no migration is needed when switching between the two).

gedankenstuecke:
That's amazing, thanks @llewelld!

mastoffel (approved):
This is excellent and works really well. I've got a few high level comments (nothing of great importance):

- When moderators reply to a comment, it feels like the story should not be able to be "in review" or "not reviewed", as they would land in "Pending Review" on the my_stories page, and the user might therefore not be aware of the comments. We could have a pop-up saying that with a comment, the moderator has to set the dropdown to "rejected".
- When the moderator wants to change their comments, there isn't an edit button but they would rather "resend" the moderated story with a new comment, which then replaces the old comment for the user. It might be good to have an indicator (a pop-up again?) to tell the moderator explicitly that the old comment will be replaced and the users won't see it anymore (as I would wonder whether the user sees a "history" too).
- This isn't anything immediate, but just a thought: it feels like the user might want to respond to the comments too, in case something has been misunderstood.
- Another thought: In case stories have longer histories (which might not be the case), it might be good to sort the experiences table with the newest events first.

gedankenstuecke:
Just converted this and its sibling PR to drafts as they are open for discussion :)

llewelld:
Thanks for your feedback @mastoffel, that's really helpful. Here are a few responses:

> * When moderators reply to a comment, it feels like the story should not be able to be "in review" or "not reviewed", as they would land in "Pending Review" on the my_stories page, and the user might therefore not be aware of the comments. We could have a pop-up saying that with a comment, the moderator has to set the dropdown to "rejected".

This is a nice point. I'll amend this PR to add in the block as you describe. I'm thinking it would also make sense to save the comment (but not share it with the author) while the story is in the 'in review' state. This would be helpful in case a moderator wants to pause their review mid-way through.

> * When the moderator wants to change their comments, there isn't an edit button but they would rather "resend" the moderated story with a new comment, which then replaces the old comment for the user. It might be good to have an indicator (a pop-up again?) to tell the moderator explicitly that the old comment will be replaced and the users won't see it anymore (as I would wonder whether the user sees a "history" too).

Perhaps the author _should_ be able to see a history of previous comments, at least until the story is accepted?

I'm struggling to think what the right flow should be here. Having a pop-up that's guaranteed to fire every time doesn't seem ideal to me. Perhaps the review comment field should be filled out with the comment from the previous review, so that the moderator can edit it rather than write a new one from scratch each time?

> * This isn't anything immediate, but just a thought: it feels like the user might want to respond to the comments too, in case something has been misunderstood.

Yes, if I recall correctly this is something that was discussed at the review meeting on the 21/4/2023 and again during the [May Meetup](https://docs.google.com/document/d/1mYcTs0kfh8HWK1X87Wc5whARrhH_Rm9xaw2a-uT6UFs/edit#heading=h.caqz9vddyyxh) (see "Adding in ability for moderators to comment on story"). Unfortunately I can't find the specifics. It wouldn't be hard to add this feature, but my recollection is there were some reservations about including it.

gedankenstuecke:
@llewelld Following the community call: shall we close down this PR in favor of #494?

llewelld:
Agreed @gedankenstuecke, I'll close this now. I've moved @mastoffel's comments to issue #525 to avoid them getting lost.

PR #510, opened 2023-05-22
Add user profile page

llewelld:
Adds an initial placeholder user profile implementation. The profile fields are just examples at this stage. The profile page isn't yet linked in to the menu, you'll need to visit the page by entering its url as <site>/users/profile/

The current implementation adds a new app "users" which stores profile data linked to the logged in User instance. It can store and display the data but not much else.

On initial integrating of these changes, you'll need to migrate the database using the following:

```
docker exec -it autspaces-web-1 python manage.py migrate users
```

After rolling back these changes you should reverse the migration with the following:

```
docker exec -it autspaces-web-1 python manage.py migrate users zero
```


server/apps/users/forms.py
```diff
@@ -0,0 +1,61 @@
+from django import forms
+
+class UserProfileForm(forms.Form):
+
+    new_user = forms.BooleanField(label = 'Is this your first time accessing the site?', required=False)
```
gedankenstuecke, line 5: I guess this shouldn't necessarily be a question the user is asked, but rather handled internally where we know whether the account was just freshly created or not (e.g. by taking advantage of the django signals). 

Similarly, we can keep internally track whether a person has ever submitted the form/edited their profile responses to keep track of whether we should give a nudge to fill those.

server/apps/users/helpers.py
```diff
@@ -0,0 +1,34 @@
...
+    profile.age_bracket = data["age_bracket"]
+    profile.gender = data["gender"]
+    profile.self_identification = data["self_identification"]
+    profile.description = data["description"]
+    #profile = UserProfile(user=user)
+
+    # .save() updates if primary key exists, inserts otherwise.
+    profile.save()
+
+def get_or_create_user_profile(user):
```
gedankenstuecke, line 24: This is a bit confusing to me, why do we need to make a `get_or_create` function when we then just call the internal one? I guess it might just be a matter of naming this function differently, as I suspect it's designed to process the data rather than actually creating the user?

gedankenstuecke (commented):
Overall no strong feelings from my side on the app-structure (ie. whether we should break this out into its own app), mainly as there's so many conflicting suggestions and recommendations that I strongly suspect it's a matter of personal preference 😂 

Otherwise just a few comments which might be too early/nitpicky at this draft/WIP stage!

server/apps/users/forms.py
llewelld: Thanks, yes, you're right, all of this functionality is due to be hidden from the user. At this stage this is just here to ensure there's a field available in the database table.

server/apps/users/helpers.py
llewelld: Thanks! I'm hoping this can be simplified out as it's just marshalling the data, but I'll bear what you say in mind and give it a clearer name in case it ends up staying.

llewelld:
I've updated the fields in preparation for the meeting tomorrow. Also rebased on main.

llewelld:
For reference, in relation to migrations. To list all the migrations:
```
docker exec -it autspaces-web-1 python manage.py showmigrations
```
To apply the migrations for this PR:
```
docker exec -it autspaces-web-1 python manage.py migrate users
```
To remove the migrations when reverting back to main branch:
```
docker exec -it autspaces-web-1 python manage.py migrate users zero
```
I'll squash the migrations and commits before taking this out of draft.

server/apps/users/helpers.py
llewelld: This code has now been removed - it wasn't needed, as you suspected.

llewelld:
I've squashed the commits, squshed the migrations and rebased, and so this is ready for review now.

server/apps/users/models.py
```diff
@@ -0,0 +1,27 @@
...
+    Holds profile information about a user
+    """
+    user = models.OneToOneField(User, on_delete=models.CASCADE)
+    profile_submitted = models.BooleanField(default=False)
+    age_bracket = models.TextField(default="unspecified")
+    gender = models.TextField(default="unspecified")
+    self_identification = models.TextField(default="unspecified")
+    description = models.TextField(default="")
+    location = models.TextField(default="")
+    comms_review = models.BooleanField(default=False)
```
gedankenstuecke, line 15: Probably something to discuss with the community what the default on this should be. I'm worried that people will never get feedback if the default is not receiving and people might click away the form but of course opting-in is good practice.

server/apps/main/views.py
```diff
@@ -118,11 +124,23 @@ def overview(request):
         context = {
             "oh_id": oh_member.oh_id,
             "oh_member": oh_member,
+            "oh_user": oh_member.user,
             "oh_proj_page": settings.OH_PROJ_PAGE,
         }
         return render(request, "main/home.html", context=context)
     return redirect("index")
 
+def login_user(request):
```
gedankenstuecke, line 133: yep, think that's a great solution!

server/apps/users/helpers.py
```diff
@@ -0,0 +1,37 @@
...
+    the profile page (because they might legitimately want the profile to be
+    emtpy).
+
+    Args:
+        user: request.user
+
+    Returns:
+        True if the user has ever submitted a profile update, False o/w.
+    """
+    profile, created = UserProfile.objects.get_or_create(user=user)
```
gedankenstuecke, line 35: I guess this function isn't used just yet, but maybe worth to check whether we need to do the `get_or_create` here (as mentioned above)

server/apps/users/views.py
```diff
@@ -0,0 +1,47 @@
...
+                UserProfile.objects.update_or_create(user=request.user, defaults=form.cleaned_data)
+
+            return redirect("users:profile")
+        else:
+            first_visit = not user_profile_exists(request.user)
+            profile, created = UserProfile.objects.get_or_create(user=request.user)
+            data = model_to_dict(profile)
+            form = UserProfileForm(data)
+            oh_member = request.user.openhumansmember
+            return render(
```
gedankenstuecke, line 38: I think the `title` context is missing to make sure the template renders the part after _AutSPACEs –_ as well

server/apps/users/views.py
```diff
@@ -0,0 +1,47 @@
...
+
+            if form.is_valid():
+                # Update the status to indicate the user has submitted at least once
+                form.cleaned_data["profile_submitted"] = True
+                UserProfile.objects.update_or_create(user=request.user, defaults=form.cleaned_data)
+
+            return redirect("users:profile")
+        else:
+            first_visit = not user_profile_exists(request.user)
+            profile, created = UserProfile.objects.get_or_create(user=request.user)
```
gedankenstuecke, line 34: Same question here: Do we need both the `get_or_create` and the `update_or_create`?

gedankenstuecke (commented):
Great work @llewelld, it works like a charm and the implementation is super reasonable as well! I just left a few comments here and there where things might be simplified a bit!

server/apps/users/forms.py
```diff
@@ -0,0 +1,105 @@
...
+    ]
+    gender = forms.ChoiceField(choices = genders, widget = forms.Select(), required=False)
+    gender.group = 1
+
+    # Taken as examples from:
+    # https://icd.who.int/browse10/2015/en#/F84
+    # See also:
+    # https://www.autism.org.uk/advice-and-guidance/topics/diagnosis/diagnostic-criteria/all-audiences
+    # Not intended as final
+    self_identifications = [
```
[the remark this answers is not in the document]
GeorgiaHCA: https://journals.sagepub.com/doi/abs/10.1177/1362361315588200 - this article is useful as it gives advice on terminology based on consultation with autistic people. "autistic person" is preferred over "person with autism". "high" and "low" functioning are also often used to differentiate autistic people on their level of care needs. This article is useful and recommends referring "high", "medium" and "low" support needs instead: https://themighty.com/topic/autism-spectrum-disorder/autism-functioning-labels-low-functioning-high-functioning/. Also many people realise or suspect they are autistic without having a formal diagnosis.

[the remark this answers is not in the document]
gedankenstuecke: Thanks @GeorgiaHCA! I would suggest we start with one or two closed dropdown questions: 

1. Do you identify as autistic: Yes/No
2. Were you diagnosed as autistic?: Not yet / Yes, as a child / Yes, as an adult 

And then we have something quite simple to request feedback on next week?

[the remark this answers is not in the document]
GeorgiaHCA: We also need to think about how we ask about connection to autism. It would be useful to know if someone is also a parent/relation/carer of an autistic person or works with autistic people professionally.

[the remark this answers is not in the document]
gedankenstuecke: @GeorgiaHCA Thanks, let's maybe put those suggestions into the meeting agenda for next week to discuss there?

server/apps/users/models.py
llewelld: Understood. I've left it as it is for now, but ready to change it after discussion. If you'd prefer us to present it with the default set to True, let me know and I can switch it.

server/apps/main/views.py
```diff
@@ -305,11 +323,16 @@ def list_public_experiences(request):
         }
     else:
         # Check the allowed triggers
-        allowed_triggers = request.GET.keys()
-        
+        allowed_triggers = set(request.GET.keys())
+
+    if "searched" not in request.GET:
+        profile, created = UserProfile.objects.get_or_create(user=request.user)
```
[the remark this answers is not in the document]
llewelld: You're right, it shouldn't have been potentially writing to the database on a GET request. I've changed it to avoid this.

server/apps/users/forms.py
```diff
@@ -0,0 +1,105 @@
...
+        widget=forms.CheckboxInput(attrs={"class": "custom-control-input"}))
+    comms_review.group = 3
+
+    # New user management
+    profile_submitted = forms.BooleanField(label = "Profile submitted",
+        required=False,
+        widget=forms.CheckboxInput(attrs={"class": "custom-control-input"}))
+    profile_submitted.group = "hidden"
+
+    def __init__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
llewelld: As I understand it (after taking a look at [the docs](https://docs.djangoproject.com/en/4.2/ref/forms/api/#initial-form-values)) the defaults won't come from the model automatically (the two are independent of one another). You can pass in an `initial` dictionary, or bind the form to the values from a model instance though. Nevertheless in this case the `__init__()` wasn't needed so I've removed it and I've moved the initial values into the field declarations instead.

server/apps/users/helpers.py
llewelld: You're right, it's currently unused (but intended for use in case we want to prod the user to complete the form at a later date). I've switched it to just use `get`.

server/apps/users/views.py
llewelld: Thank you! I've given it a title now.

server/apps/users/views.py
llewelld: Thanks for flagging these as well. I've removed the write in the case of the `GET`. For the `POST` a missing `UserProfile` shouldn't happen in theory, but django forces us to condition on the possibility, and I'm not sure there's a more sensible approach than creating the profile if there isn't one.

server/apps/users/forms.py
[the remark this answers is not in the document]
llewelld: Thanks for the feedback on this. I've updated the field to simplify it basd on the discussion, and so that it can be considered at the next meetup. I've also created task #533 for updating the fields once they've been finalised.

server/apps/users/views.py
gedankenstuecke: I guess we can figure out how to refactor this through some custom decorator that would redirect or something later on!

I guess my main concern is the following path: 
1. `UserProfile` doesn't get created during/after login 
2. One of the error-catching `get_or_create` produces the profile
3. The profile now exists, but the user never sees the profile editing page, as they weren't directed to it and don't know it exists

If the "default" behaviour of redirecting people to the profile form after first logging in were to break for some reason, we'd a) not know as it never leads to outright errors and b) now all users are going down the path I outlined above? 

But I might be overthinking the danger of this and I'm sure we'll touch these flows again as part of the user testing, so let's not be the perfect the enemy of the good, i.e. block this PR on sorting a better approach out

server/apps/users/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.urls import path, re_path
+from server.apps.users import views
+
+app_name = "servers.apps.users"
+
+urlpatterns = [
+    path("profile/", views.user_profile, name="profile"),
+    path("greetings/", views.user_profile, {"first_visit": True}, name="greetings"),
```
gedankenstuecke, line 8: This might be a left-over from testing? 🙂

server/apps/users/models.py
gedankenstuecke: Let's leave it like this for now and we ask for feedback on it during the community call? :)

server/apps/users/forms.py
```diff
@@ -0,0 +1,91 @@
...
+
+    genders = [
+        ("nonbinary", "Non-binary"),
+        ("female", "Female"),
+        ("male", "Male"),
+        ("see_description", "See description"),
+        ("unspecified", "Prefer not to say"),
+    ]
+    gender = forms.ChoiceField(choices = genders, widget = forms.Select(), required=False,
+                              initial="unspecified")
```
gedankenstuecke, line 26:
```suggestion
    gender = forms.ChoiceField(choices = genders, widget = forms.Select(), required=False,
                              label="Which gender do you identify as?",
                              initial="unspecified")
```

server/apps/users/forms.py
```diff
@@ -0,0 +1,91 @@
...
+        ("under 18", "Under 18"),
+        ("18-25", "18-25"),
+        ("26-35", "26-35"),
+        ("36-45", "36-45"),
+        ("46-65", "46-65"),
+        ("over 65", "Over 65"),
+        ("unspecified", "Prefer not to say"),
+    ]
+    age_bracket = forms.ChoiceField(choices = age_brackets, widget = forms.Select(), required=False,
+                                   initial="unspecified")
```
gedankenstuecke, line 15:
```suggestion
    age_bracket = forms.ChoiceField(choices = age_brackets, widget = forms.Select(), required=False,
                                   label="What is your age?",
                                   initial="unspecified")
```

server/apps/users/forms.py
```diff
@@ -0,0 +1,91 @@
...
+    # Not intended as final
+    autistic_identifications = [
+        ("yes", "Yes"),
+        ("no", "No"),
+        ("not_sure", "Not sure"),
+        ("self_describe", "Prefer to self-describe"),
+        ("unspecified", "Prefer not to say"),
+    ]
+    autistic_identification = forms.ChoiceField(choices = autistic_identifications, widget = forms.Select(),
+                                           required=False, initial="unspecified")
```
gedankenstuecke, line 38:
```suggestion
    autistic_identification = forms.ChoiceField(choices = autistic_identifications, widget = forms.Select(),
                                           label="Do you identify as autistic?",
                                           required=False, initial="unspecified")
```

server/apps/users/forms.py
```diff
@@ -0,0 +1,91 @@
+from django import forms
+
+class UserProfileForm(forms.Form):
```
gedankenstuecke, line 3: Suggested adding custom labels on the initial demographic questions to clarify what is asked compared to defaults :)

gedankenstuecke (approved):
I think from my side this looks like an amazing MVP of the `UserProfile` approach, thanks so much @llewelld! 

I'd be happy to have this being merged in if that's what you'd like. Otherwise we could also wait for the feedback next week if you prefer keeping improvements to be part of this PR? Really up to you @llewelld! 😃

server/apps/users/views.py
llewelld: Thanks for clarifying. What would be your prefered behaviour in case this situation arises?

For info, I've included unit tests to try to catch this, but that's no guarantee, and I can envisage situations when it might happen (e.g. if the `OPENHUMANS_LOGIN_REDIRECT_URL` were to change).

This could also, to some extent, be mitigated if we end up adding in the funcitonality to occassionally remind the user in case they've never submitted a profile.

llewelld: Also, thinking about it further, the situation here only arises if the user is actually submitting their profile, so perhaps in this case it's not such an issue?

gedankenstuecke: Yeah, I think in a way the preferred behaviour would be to check whether the `UserProfile` exists in a consistent way across all views and if not only create it but also redirect the user to the profile entry page?

gedankenstuecke: And agree, for the POST of the UserProfile form it makes sense 👍

server/apps/users/urls.py
llewelld: This is the fix I added to avoid using `UserProfile` existence as a "first visit" condition. The OpenHumans redirect page is now the greetings page, which happens to be the profile page + greeting message.

This is definitely preferable to the approach used earlier, because it avoids writing on `GET` of the profile page, which I understand is generally frowned upon.

server/apps/users/forms.py
llewelld: Thanks!

server/apps/users/urls.py
gedankenstuecke: Ah, got it, but does that mean that the `login_user` in the `main.views` is now unused and that by default no `UserProfile` is created when logging in?

llewelld: Ah, no, that's my mistake in how I described it (sorry). To correct what I wrote: `login_user` is the entry point. That then either redirects to `greetings` or `overview` depending on whether it's the first visit or not.

gedankenstuecke: No worries at all! Web stuff is always such a complex beast! :D 

Last double check that I got it, the flow is: `login_user`, which creates the `UserProfile` and then we move the user to `greetings`?

server/apps/users/views.py
llewelld: My feeling is this needs a bit more thought (I've created #534 to capture those thoughts). Would you mind if we leave it like this for now and revisit it in a seprate task?

gedankenstuecke: Yes, absolutely!

llewelld:
Thanks for the helpful feedback. My preference would be to merge it in now (once it's at a state you're happy with), and then create new PRs for subsequent changes after the meeting. The post-meeting discussion is likely to result in database changes, so I understand if you prefer to wait until they're finalised.

server/apps/users/urls.py
llewelld: That sounds correct to me, yes (the [relevant code](https://github.com/llewelld/AutSPACEs/blob/9782fbe132e8c8fbcce26bd914edc1fc115875d9/server/apps/main/views.py#L133)).

gedankenstuecke: Yep, that's great! Thanks for walking me through it 🙏

llewelld: (just noticed: I will fix the spelling in that docstring!)

gedankenstuecke:
I'd be happy to have it merged in, feel free to push that huge green button 😄

Issue #497, opened 2023-05-03
Specify behaviour when form is not valid - views `share_experience.py`


helendduncan:
### Summary

There is no "else" to the if form.is_valid() in `share_experience` function in `views.py`
This means invalid forms (eg character length too long see #491) you will get a None Http response and an error

### Deadline 

May 2023

### What needs to be done?

Decide on sensible action for invalid form (back to the edit page perhaps?) and implement it

### Who can help?

* @helendduncan 
* 

---

gedankenstuecke:
Great catch! I would say returning the user to the edit/enter page would be the best. I think by re-using the invalid `form` this should also include the error message we can display on the frontend side alongside the existing data!

gedankenstuecke:
Closed by #512

Issue #491, opened 2023-04-24
Forms: increase the character limit


helendduncan:
### Summary

Currently there is a 500 character limit in forms.py for the experience and difference text and a 150 character limit for the title text. This is not long enough for the current placeholder stories that we are using.

Note there is no limit in the models, just the form

Need to decide if to place a limit on the number of characters and if so what it should be.

### Deadline 

May 2023

### What needs to be done?

- [X] Decide new character limit - "Very large" - HD suggests 10,000 characters
- [ ] Implement change (have done it locally and works)

### Who can help?

* @gedankenstuecke 
* @GeorgiaHCA 

---

### Updates

4th May 2023:
Cut and paste from [Google doc](https://docs.google.com/document/d/1mYcTs0kfh8HWK1X87Wc5whARrhH_Rm9xaw2a-uT6UFs/edit#heading=h.caqz9vddyyxh) from today's monthly meet up
We’ll begin by implementing no or a very large word [character] count for the MVP, and do user testing and iterate if people send in unworkable large experiences.
Titles should be consistent and readable and fitted to user screen size - this is the most important part to get right- we should also impose a length and cut off rest from story - but don’t cut off title.

HD: In practice I think setting a 10,000 character limit for the experience and the same for the recommendation would be a sensible for the first iteration. This approximately corresponds to three sides of A4 with 12 point font (and some line/paragraph breaks)
It has been suggested by @gedankenstuecke that once we have users interacting with the platform we can analyse story lengths and modify any limits based on that.

HD: Currently there is a 150 character limit for the title text. RE: limiting length will either remove cap or set it to be 10,000 characters. RE: viewing. Currently the title remains fully visible for rescaling of the page in my test environment (using Chrome) may want to think of a systematic way to monitor these visual requirements?

#### Some selected relevant comments from 4th May meet up:
* Feedback from a community member: it can be distressing to have a limit and count down. Response: Count up would be better than count down
* From a psychological perspective it might put people off if there is a cap - some autistic people may like to submit a large amount of text - this is part off accessibility to include autistic people’s preferred mode of expression
* You shouldn’t allow more than a moderator could reasonably moderate
* Is there a number of characters after which the software breaks?
* Could you chunk up the text for the moderators, while giving the users a more seamless experience
* MVP should be particularly focussed on being sensitive to user needs so people aren’t put off on the initial stages
[...]

PR #512, opened 2023-06-12
Fix experience submission bugs

gedankenstuecke:
This PR fixes some smaller and bigger issues: 

- #497: So far the `share_experience` would fail by not returning any `HttpResponse` when the experience form was invalid. This is now fixed, instead it re-routes back to the form with all the data the user had entered so far, giving them a chance to edit the submission
- So far we didn't use the `alert` `messages` of Django at all, despite it being part of the `application.html`. I've cleaned those up a bit and made them functional. When an invalid form is returned, it adds the `warning alert` to highlight the user about what needs to be edited (see screenshot below) 
- #491: removes the character limits for the experience & difference texts

Screenshot 2023-06-12 at 16 42 11

@llewelld @helendduncan I know you're at your hack week this week, but once you're back it would be amazing if one of you could review those changes 🙏


gedankenstuecke:
Added a test to cover the new lines re: gracefully failing and also removed print debug-statements 😂

gedankenstuecke:
And caught the last 2 stray lines 😉

server/apps/main/views.py
```diff
@@ -150,7 +152,28 @@ def share_experience(request, uuid=False):
...
+                    moderation_status = "not reviewed"
+                    title = "Share experience"
+                return render(
+                    request,
+                    "main/share_experiences.html",
+                    {
+                        "form": form,
+                        "uuid": uuid,
+                        "title": title,
+                        "moderation_status": moderation_status,
```
mastoffel, line 174: David pointed me to there not being a `viewable` variable here.

However, I think the share_experience.html template also doesn't make of the `viewable` variable, although it is returned later in this function (line 201). Should we delete it later on or add it here?

static/css/main.css
```diff
@@ -151,6 +151,11 @@ body {
  }
 
  /*Shared Experience Section*/
+
+ .alert-container {
+  padding-top: 10px;
+ }
```
mastoffel, line 157: The alerts are slightly less wide than the containers directly below, maybe it would be possible to have them the same width?

mastoffel (commented):
Great work Bastian! Leaving any or all of the text boxes out gives the correct messages. 

One high level comment:
My first feeling is that it should be ok to leave "What could have made your experience better?" blank, as this seems like it would require quite a bit of additional effort after writing the story, and doesn't seem crucial. But not sure whether it is necessary for the research side.

server/apps/main/tests/test_views.py
```diff
@@ -231,6 +231,55 @@ def test_share_exp_submit_new_experience(self):
...
+        user_a_stories_after = len(
+            PublicExperience.objects.filter(open_humans_member=self.oh_a)
+        )
+
+        pe_db_after = PublicExperience.objects.filter(title_text="A new story added")
+
+        # Check that a story with the title now exists in the database
+        assert len(pe_db_after) == 0
+
+        # Check that there is one new story for user A
```
mastoffel, line 275:
```suggestion
# Check that is no new story for user A
```

server/apps/main/views.py
gedankenstuecke: I noticed that as well! I think this is a result of how the form later on (in the `else` part for non-POSTs) is being initialized by being a merge from the `PublicExperience` object and the OH file. 

So I wasn't 100% sure if we could remove it there as well - but we probably could, but maybe worth not putting this into the PR here but rather clean up as its own PR? :)

mastoffel: sounds good!

static/css/main.css
mastoffel: ah yes, that makes sense!

server/apps/main/tests/test_views.py
```diff
@@ -231,6 +231,55 @@ def test_share_exp_submit_new_experience(self):
...
+        user_a_stories_before = len(
+            PublicExperience.objects.filter(open_humans_member=self.oh_a)
+        )
+
+        response = c.post(
+            "/main/share_exp/",
+            {
+                "experience_text": "",
+                "difference_text": "",
+                "title_text": "",
```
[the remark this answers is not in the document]
gedankenstuecke: Added some assertions to test that it gives back errors for all three!

gedankenstuecke:
Thanks @mastoffel! I think the high-level question for which aspects should be mandatory probably requires a conversation and I wouldn't go ahead editing it now :)

PR #513, opened 2023-06-13
fix button declarations

gedankenstuecke:
On the post-experience-submission confirmation page there's a number of buttons for next steps. These were somehow weirdly nested between `<a href>` and `<button>` stylings, making them appear slightly strange and less accessible through keyboard navigation (e.g. via `vimium`). 

I've just gone ahead and fixed it :D


mastoffel (commented):
Works well and keyboard navigation is smooth now!

gedankenstuecke:
Thanks!

PR #519, opened 2023-06-21
Clean experience submission view/template

gedankenstuecke:
Thanks to @mastoffel's feedback on #512 I had a look over the `share_experience` view and template and fixed some more minor things: 

1. I removed the unused `viewable` variable that was passed on into the template but wasn't used anywhere
2. There was another minor bug in which the `<label for="field_id"/>` didn't work for the public sharing option, due to it being a custom declaration to make @mastoffel's excellent JS work more nicely. I've fixed that and also unified using the right label throughout the template to avoid future issues.

### DIFF
--- a/server/apps/main/templates/main/share_experiences.html
+++ b/server/apps/main/templates/main/share_experiences.html
@@ -161,7 +161,7 @@
         {% if show_moderation_status %}
           {% for field in group.list %}
           <div class="form-group">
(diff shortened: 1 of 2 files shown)


mastoffel (commented):
Well done, definitely learned something here :)

gedankenstuecke:
Thanks for the review! The `id_for_label` is [hidden a bit in the forms API](https://docs.djangoproject.com/en/4.2/ref/forms/api/) (linking for future reference) 🙈

PR #522, opened 2023-06-21
Single story view option

### LINKED ISSUE #498: View a single story on a single page
### Summary
Following a comment at the May meet up (Google doc [here](https://docs.google.com/document/d/1mYcTs0kfh8HWK1X87Wc5whARrhH_Rm9xaw2a-uT6UFs/edit#heading=h.caqz9vddyyxh)):
Would it be possible, when you click “read the story”, will it remain on the page or go onto another page - if there’s one immediately after, it may be too much, and having it separate would be easier to share.

> Another page might make it easy to copy and paste content and share a link with others/bookmark a story that you want to come back to

> Can the default be the infinite scroll on the expanded story be an option to “open in a new page” 

> Adding a button of “open in a new page”

> “Continue reading?” button/option after a certain cut off so that you are not bound on continuing 

> Having all on one page might clutter the experiences page 

> Reactive to screen width - we can alter text display based on size of screen 

> There is a minimum recommended font size - make sure it doesn’t drop below that font size.
(see issue #490)

### Deadline 
June 2023
### What needs to be done?
- [ ] Make template for single story page
- [ ] Add "read as standalone" button to accordion (? once opened or when closed ?)
- [ ] Add a views function to send contents of story with given uuid to the template

### Who can help?

* 
* 

---

helendduncan:
Closes #498 

Single story page with formatting similar to the moderation page

As well as feedback on technical features - any feedback/edits about layout greatly appreciated

### DIFF
--- a/server/apps/main/templates/main/single_story.html
+++ b/server/apps/main/templates/main/single_story.html
@@ -13,7 +13,13 @@
   <div class="container story-section">
 
     {% for experience in experiences %}
+    <div class="container story-section">
+    {% if experience.abuse or experience.violence or experience.drug or experience.mentalhealth or experience.negbody or experience.other %}
+    <span class="badge badge-warning float-right" style="font-size: 1rem; margin-top:2%">Includes triggering contents</span>
+    {% endif %}
+    </div>
 
+    <div class="container story-section">
     <h2>Title: <i>{% firstof experience.title_text "No Title Given" %}</i></h2>
 
     <h3>Experience</h3>
@@ -21,6 +27,7 @@
 
     <h3>What could have made the experience better?</h3>
     <p><i>{% firstof experience.difference_text "No Difference Text Given" %}</i></p>
+    </div>
 
     {% endfor %}
   </div>
--- a/server/apps/main/templates/main/single_story.html
+++ b/server/apps/main/templates/main/single_story.html
@@ -1,6 +1,6 @@
 {% extends 'main/application.html' %}
 
-{% block title %}AutSPACEs - {{title}} {% endblock %}
+{% block title %}AutSPACEs - {{ title }} {% endblock %}
 
 {% load static %}
 {% load custom_tags %}
--- a/server/apps/main/templates/main/single_story.html
+++ b/server/apps/main/templates/main/single_story.html
@@ -1,0 +1,31 @@
+{% extends 'main/application.html' %}
+
+{% block title %}AutSPACEs - {{title}} {% endblock %}
+
+{% load static %}
+{% load custom_tags %}
+{% load humanize %}
+
+{% block content %}
+
+<section id="single_story">
+
+  <div class="container story-section">
+
+    {% for experience in experiences %}
+
+    <h2>Title: <i>{% firstof experience.title_text "No Title Given" %}</i></h2>
+
+    <h3>Experience</h3>
+    <p><i>{% firstof experience.experience_text "No Experience Text Given" %}</i></p>
+
+    <h3>What could have made the experience better?</h3>
+    <p><i>{% firstof experience.difference_text "No Difference Text Given" %}</i></p>
+
+    {% endfor %}
+  </div>
+
+
+</section>
+
+{% endblock %}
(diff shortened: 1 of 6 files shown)


server/apps/main/views.py
```diff
@@ -502,3 +508,20 @@ def moderate_experience(request, uuid):
             )
     else:
         return redirect("index")
+
+
+def single_story(request, uuid):
```
gedankenstuecke, line 513: I guess we always expect there to be only a single experience so the `filter` and passing a list might be a bit overkill. 

A simpler solution could be a try/except: 

```
try:
  experience = PublicExperience.objects.get(experience_id=uuid,moderation_status='approved')
  … make context
  return render(…)
except DoesNotExist:
  return redirect(…)
```

That way we don't have to check for the length etc. and have a slightly clearer code?

gedankenstuecke (commented):
Works great, just added one potential edit to the actual view to simplify querying for the story 🙂

server/apps/main/views.py
helendduncan: Thats nice and neat! Will implement

server/apps/main/views.py
```diff
@@ -11,6 +11,7 @@
 from django.db.models import Q
 from django.urls import reverse
 from django.core.paginator import Paginator
+from django.core.exceptions import EmptyResultSet, ObjectDoesNotExist
```
gedankenstuecke, line 14:
```suggestion
from django.core.exceptions import ObjectDoesNotExist
```

In the original PR doesn't need either as unused. With my suggested edit (below), need the `ObjectDoesNotExist` exception

server/apps/main/views.py
```diff
@@ -502,3 +509,22 @@ def moderate_experience(request, uuid):
...
+        experiences = PublicExperience.objects.filter(
+            experience_id=uuid, moderation_status="approved"
+        )
+        title = experiences.values_list("title_text", flat=True)[0]
+        exp_context = {"experiences": experiences}
+        title_context = {"title": title}
+        context = {**exp_context, **title_context}
+        return render(request, "main/single_story.html", context=context)
+    except:
+        return redirect("main:overview")
```
gedankenstuecke, line 530:
```suggestion
    try:
        experience = PublicExperience.objects.get(
            experience_id=uuid, moderation_status="approved"
        )
        title = experience.title_text
        context = {
            "experience": experience,
            "title": title
            }
        return render(request, "main/single_story.html", context=context)
    except ObjectDoesNotExist:
        return redirect("main:overview")

```

This would be my suggestion for how to handle it, it tries to `get` the single experience based on the ID & moderation status and render the HTML. If it excepts because of the object not existing redirect. 

This change also handles a single experience to the template for rendering, so could remove the for-loop in the template (see above).

server/apps/main/templates/main/single_story.html
```diff
@@ -0,0 +1,40 @@
...
+{% load custom_tags %}
+{% load humanize %}
+
+{% block content %}
+
+<section id="single_story">
+
+  <div class="container story-section">
+
+    {% for experience in experiences %}
```
gedankenstuecke, line 15:
```suggestion

```

server/apps/main/templates/main/single_story.html
```diff
@@ -0,0 +1,40 @@
...
+    <h2>Title: <i>{% firstof experience.title_text "No Title Given" %}</i></h2>
+
+    <h3>Experience</h3>
+    <p><i>{% firstof experience.experience_text "No Experience Text Given" %}</i></p>
+
+    <h3>What could have made the experience better?</h3>
+    <p><i>{% firstof experience.difference_text "No Difference Text Given" %}</i></p>
+    </div>
+
+    {% endfor %}
```
gedankenstuecke, line 34:
```suggestion

```

gedankenstuecke (commented):
I tried simplifying it a bit more to avoid having to fetch and loop over a queryset. :)

gedankenstuecke:
Thanks @helendduncan, that looks great! 👍

helendduncan:
Happy for a merge @gedankenstuecke?

PR #526, opened 2023-06-23
Explicitly state which trigger warnings are selected for a story

helendduncan:
I honestly thought there was an issue for this but I can't see it - so here is a proof of concept - perhaps @gedankenstuecke can let me know if it is needed or not,
Screenshot 2023-06-23 at 22 04 07


gedankenstuecke (commented):
Thanks @helendduncan, I could have sworn as well that we had an issue for adding the actual labels to the experiences. I guess this might have been discussed in one of the broader "filter by trigger labels" issues that were closed by your filtering PR! 

But in any case, this looks great so far! I think there's just two minor, potential improvements that would come to mind: 

1. Do we want the `other` trigger label just say other or replace it by `other: $user_text`? Or use `Other trigger warning`? Just using _Other_ might be a bit confusing. Providing the actual rigger warning would be useful but could potentially be too long and break the UI? 
2. Currently the "badges" don't have any separation between them based on the screenshot. Could we spread them out slightly so that they don't run into each other? :)

helendduncan:
Current version: If all checkbox-triggers are clicked the "all trigger warnings selected" is shown. If all checkboxes plus "other" are picked then then the text of the other tag is shown.

It is trivial to go to "Other trigger warning" instead if this is too cumbersome

No limit on the length of the other tag has been built in if limiting it suggest it happens at the model/form stage? Also might need to decide on behaviour for multiple other tags but that is likely beyond the scope of this PR

Screenshot 2023-06-26 at 17 02 21

gedankenstuecke (approved):
Looks great, thanks @helendduncan 👍

helendduncan:
@mastoffel can you have a quick check that this still behaves as expected with pagination - It looks sensible to me (aside from url....) but want to check before merge - Cheers!

PR #529, opened 2023-06-27
Paginate overview

mastoffel:
Paginates the overview page. 

- keeps toggles
- continuous numbering across pages


gedankenstuecke:
I tried resolving the merge conflict. @mastoffel maybe you can have a look if that all looks reasonable and otherwise we can go and merge I think!

mastoffel:
Thanks Bastian! Yes, agree with url, let's see whether this automatically getting fixed with the _all_ toggle.

I tried it again and things are looking good, thanks for doing the merge!

PR #532, opened 2023-06-29
Mystories numbering

### LINKED ISSUE #455: Add pagination to my_stories, shared_stories pages, and search results

### LINKED ISSUE #530: Numbering of stories on the "mystory" page
### Summary

Currently, when there are more than 10 stories and the pagination splits them across multiple pages, the numbering starts at 1 again at every page. It would be better to have continuous numbering across pages.

### What needs to be done?

- Implement continuous numbering for stories in each tab

mastoffel:
* Continuously number my_stories within each tab / category
* Write proper tests for pagination / numbering
* my_stories and the experiences overview page now use the same pagination and numbering functions

Thanks @helendduncan for all the help and support! 

Closes #530
Closes #455

P.S. hopefully nothing too Mysterious about this (except the cassettes)


gedankenstuecke:
I totally read that PR title as _mysterious numbering_ and was wondering what that was all about! 😂

server/apps/main/tests/test_views.py
```diff
@@ -672,8 +672,63 @@ def test_list_public_exp_pagination(self):
...
+    # 2 are in review
+    # 22 are not reviewed
+    # 25 are approved
+    # 7 are rejected
+    def test_my_stories_pagination_page1(self):
+        c = Client()
+        c.force_login(self.user_b)
+        
+        response = c.get("/main/my_stories/")
+        num_items_per_page = 10
```
gedankenstuecke, line 699: Are those breaks hardcoded in the cassette or would it work with any number? I.e. if we were to change this to 5 here, would this still work? If it would still work – should we grab the "correct" number from the settings?

server/apps/main/tests/test_views.py
```diff
@@ -672,8 +672,63 @@ def test_list_public_exp_pagination(self):
...
+            filter_query_parameters=["access_token"],
+            match_on=["path"],
+    )
+    # Same cassette as above
+    def test_my_stories_pagination_page2(self):
+        c = Client()
+        c.force_login(self.user_b)
+        # checks page 2 for public and in_review stories
+        response = c.get("/main/my_stories/?page_public=2&page_review=2")
+        num_items_per_page = 10
```
gedankenstuecke, line 724: same question as above re: `num_items_per_page`

gedankenstuecke (commented):
Overall looks great and seems to work fine on my end as well (will give it a last look tomorrow). 

But already wanted to leave one question re: the tests! :)

gedankenstuecke:
Oh, and I should say: not too mysterious so far 😂

server/apps/main/tests/test_views.py
mastoffel: Good question. I just checked and the pagination is hardcoded in the cassette. I don't really have an idea how to avoid that, but happy to hear suggestions!

gedankenstuecke: That makes sense re: the cassettes, I forgot that it needs to fetch and process the data from OH depending on where the data comes from. I guess it would work for the `PublicExperiences` view, but I think it's fine if we don't have the test adjust to any `num_items_per_page` but rather sticks with what we've set as the default!

gedankenstuecke (approved):
Just gave it one last look and all works great 👍

server/apps/main/tests/test_views.py
mastoffel: yes exactly, ok, let's stick with it for now.

PR #542, opened 2023-07-06
Updated wording in the experience page

### LINKED ISSUE #539: Update wording
### Summary

Following the July 2023 meet up ([Google doc](https://docs.google.com/document/d/1mYcTs0kfh8HWK1X87Wc5whARrhH_Rm9xaw2a-uT6UFs/edit#heading=h.7f28vh3w1oai)) the wording in two areas was mentioned.

On the public experiences page:

> We have excluded potentially triggering content, if you would like to see more content please choose which categories you would like to see

should be replaced with something not implying one is happy to see triggering content.

On the view single story button 

>Read Story on Single Page

should be replaced with something more succinct

### Deadline 
August 2023

### What needs to be done?

- [ ] Decide on wording
- [ ] Update the relevant file (server/apps/main/templates/main/experiences_page.html

### Who can help?

* @helendduncan 
* @obinnaofomah

---

obinnaofomah:
Closes #539 Just changed the wording in this page. It would be useful to get feedback or suggestions for the alternate wording.

### DIFF
--- a/server/apps/main/templates/main/experiences_page.html
+++ b/server/apps/main/templates/main/experiences_page.html
@@ -38,7 +38,7 @@
             </div>
           </div>
 
-          <div class="trigger-warning-message"> We have excluded potentially triggering content, if you would like to see more content please choose which categories you would like to see</div>
+          <div class="trigger-warning-message"> We have excluded potentially triggering content, please select any categories that you would be comfortable seeing </div>
 
 
       <div class="form-check mb-2">
@@ -177,7 +177,7 @@
           aria-expanded="false"
           aria-controls="single{{ experience.experience_id }}"
           style="text-align: right">
-          Read Story on Single Page
+          Read More
           </a>
           </div>
           </div>


server/apps/main/templates/main/experiences_page.html
```diff
@@ -177,7 +177,7 @@ <h5 class="card-title story-header">{{ experience.number }}. {% firstof experien
           aria-expanded="false"
           aria-controls="single{{ experience.experience_id }}"
           style="text-align: right">
-          Read Story on Single Page
+          Read More
```
gedankenstuecke, line 180: This one might be a bit confusing, given that the other button for expanding the is called _Read Full Story_ and already shows the whole experience. Maybe we could do the following: 

1. Rename the _other_ button from _Read Full Story_ to _Expand_
2. Rename this button to _Open in new page_

gedankenstuecke (changes requested):
The renaming on the top re: the trigger labels looks great! And I've left a suggestion of how we could rename the two buttons in question inline! 👍

gedankenstuecke (approved):
That's brilliant, thanks so much!

PR #547, opened 2023-07-10
Add warning dialogues to moderation page

### LINKED ISSUE #525: Improve moderation reply process
### Summary

@mastoffel made some very [useful comments](https://github.com/alan-turing-institute/AutSPACEs/pull/509#pullrequestreview-1434015419) on previous (now closed) PR #509. I'm logging them here as a task so that they can be acted upton.

> 1. When moderators reply to a comment, it feels like the story should not be able to be "in review" or "not reviewed", as they would land in "Pending Review" on the my_stories page, and the user might therefore not be aware of the comments. We could have a pop-up saying that with a comment, the moderator has to set the dropdown to "rejected".
> 2. When the moderator wants to change their comments, there isn't an edit button but they would rather "resend" the moderated story with a new comment, which then replaces the old comment for the user. It might be good to have an indicator (a pop-up again?) to tell the moderator explicitly that the old comment will be replaced and the users won't see it anymore (as I would wonder whether the user sees a "history" too).
>  3. This isn't anything immediate, but just a thought: it feels like the user might want to respond to the comments too, in case something has been misunderstood.
> 4. Another thought: In case stories have longer histories (which might not be the case), it might be good to sort the experiences table with the newest events first.

I think point 3 has been decided against, but the remaining three all seem valid to me and worth implementing.

### What needs to be done?

- [ ] Add pop-up to prevent incongruous moderations.
- [ ] Add pop-up to warn the user that comments will be replaced.
- [ ] Reverse the sort order of the moderation history.

### Who can help?

* Anyone who wants to add to the AutSPACEs codebase.

---

llewelld:
Under certain circumstances it may not make sense to submit a review:

1. If the story is being rejected without a reason.
2. If the status is being set to 'not reviewed' but has review comments.

This change adds a dialogue box to warn the user in the case of 1, and block the user in the case of 2.

Fixes #525.

### DIFF
--- a/server/apps/main/helpers.py
+++ b/server/apps/main/helpers.py
@@ -26,10 +26,15 @@ def is_moderator(user):
         return False
 
 
-def model_to_form(model, disable_moderator=False):
+def public_experience_model_to_form(model, disable_moderator=False):
     """Converts Model to form."""
     model_dict = model_to_dict(model)
     model_dict["moderation_prior"] = model_dict["moderation_status"]
+
+    # Copy the latest moderation reply to the form
+    if (model.experience_id):
+        change_reply, changed_at = get_latest_change_reply(model.experience_id)
+        model_dict["moderation_reply"] = json.dumps(change_reply)
 
     form = ModerateExperienceForm(
         {**model_dict, "viewable": True},  # we only moderate public experiences
@@ -592,16 +597,30 @@ def get_latest_change_reply(experience_id):
 
     return change_reply, changed_at
 
-def get_moderator_message():
-    leaf_name = "mod_message.txt"
+def get_message(file_name):
+    """
+    Loads a message to be sent to the user.
+
+    The first line of the file should contain the subject. Subsequent lines
+    will comprise the message body.
+
+    Args:
+        file_name: the leaf name of the file to load.
+    Returns:
+        subject and body of the message.
+    """
     module_dir = os.path.dirname(__file__)
-    file_path = os.path.join(module_dir, leaf_name)
-    subject = ""
+    file_path = os.path.join(module_dir, file_name)
+    subject = None
     message = ""
-    with open(file_path, "r") as f:
-        subject = f.readline()
-        for line in f:
-            message += line
+    try:
+        with open(file_path, "r") as f:
+            subject = f.readline()
+            for line in f:
+                message += line
+    except IOError as e:
+        print("Exception when opening moderation message: {}".format(str(e)))
+        message = None
     return subject, message
 
 def message_wrap(text, width):
--- a/server/apps/main/templates/main/moderate_experience.html
+++ b/server/apps/main/templates/main/moderate_experience.html
@@ -155,20 +155,20 @@
         <p><i>{{experience_difference}}</i></p>
       </div>
 
-      <template id="template-control" style="display: none;">
+      <template id="template-control" class="template-hidden">
         <div class="dropdown control">
           <button class="btn btn-secondary dropdown-toggle" type="button" id="dropdownMenuButton1" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
             Reason for moderation
           </button>
           <ul class="dropdown-menu" aria-labelledby="dropdownMenuButton1">
-            <li><a class="dropdown-item addreason moderation-red" data-reason="Includes identity or sensitive data" data-href="identity">Identity or sensitive data</a></li>
-            <li><a class="dropdown-item addreason moderation-red" data-reason="Discriminatory or belittling content" data-href="discriminatory">Discriminatory or belittling</a></li>
-            <li><a class="dropdown-item addreason moderation-red" data-reason="Includes a racial slur" data-href="racial-slurs">Racial slur</a></li>
-            <li><a class="dropdown-item addreason moderation-red" data-reason="Making assumptions when writing on behalf of others" data-href="are-you-making-assumptions">Assumptions on behalf of others</a></li>
-            <li><a class="dropdown-item addreason moderation-red" data-reason="Including your own subjective details when writing on behalf of others" data-href="are-you-adding-subjective-detail">Own subjectivity on behalf of others</a></li>
-            <li><a class="dropdown-item addreason moderation-red" data-reason="Responding to someone else's story" data-href="responding-to-others">Response to someone else's story</a></li>
-            <li><a class="dropdown-item addreason moderation-amber" data-reason="Includes swear words or disability slurs" data-href="swear-words">Swear words or disability slurs</a></li>
-            <li><a class="dropdown-item addreason moderation-amber" data-reason="Potentially triggering to some users" data-href="swear-words">Triggering to some users</a></li>
+            <li><a class="dropdown-item addreason" data-reason="Includes identity or sensitive data" data-href="identity" data-severity="red">Identity or sensitive data</a></li>
+            <li><a class="dropdown-item addreason" data-reason="Discriminatory or belittling content" data-href="discriminatory" data-severity="red">Discriminatory or belittling</a></li>
+            <li><a class="dropdown-item addreason" data-reason="Includes a racial slur" data-href="racial-slurs" data-severity="red">Racial slur</a></li>
+            <li><a class="dropdown-item addreason" data-reason="Making assumptions when writing on behalf of others" data-href="are-you-making-assumptions" data-severity="red">Assumptions on behalf of others</a></li>
+            <li><a class="dropdown-item addreason" data-reason="Including your own subjective details when writing on behalf of others" data-href="are-you-adding-subjective-detail" data-severity="red">Own subjectivity on behalf of others</a></li>
+            <li><a class="dropdown-item addreason" data-reason="Responding to someone else's story" data-href="responding-to-others" data-severity="red">Response to someone else's story</a></li>
+            <li><a class="dropdown-item addreason" data-reason="Includes swear words or disability slurs" data-href="swear-words" data-severity="amber">Swear words or disability slurs</a></li>
+            <li><a class="dropdown-item addreason" data-reason="Potentially triggering to some users" data-href="swear-words" data-severity="amber">Triggering to some users</a></li>
           </ul>
         </div>
       </template>
@@ -287,7 +287,7 @@
 
         <h5>Moderation reasons to be passed back to the author</h5>
 
-        <p>Highlight a portion of the experience text to add a reason.</p>
+        <p>Highlight a portion of the experience text to add a reason. These comments will be sent to the author on review completion.</p>
         <div id="reason-titles">
           <div class="d-flex flex-row">
             <div class="p-2 col col-lg-7 text"><b>Relevant text</b></div>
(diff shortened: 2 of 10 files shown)


llewelld:
This is interim (hence draft), there are still some further changes needed:
1. The previous moderation comments should be restored when the page is entered.
2. The form shouldn't validate if review comments are present on a 'not reviewed' story.
3. Reverse the ordering (newest first) in the experience history table.

server/apps/main/helpers.py
```diff
@@ -31,6 +31,11 @@ def model_to_form(model, disable_moderator=False):
     model_dict = model_to_dict(model)
     model_dict["moderation_prior"] = model_dict["moderation_status"]
 
+    # Copy the latest moderation reply to the form
+    if (model.experience_id):
```
gedankenstuecke, line 35: I'm not 100% sure I understand why this is necessary. Is it to handle the case where someone force submits a form (i.e. client-side JS doesn't prevent non-valid form submitted) and we need to restore the form?

server/apps/main/templates/main/moderate_experience.html
```diff
@@ -155,7 +155,7 @@ <h3>What could have made the experience better?</h3>
         <p><i>{{experience_difference}}</i></p>
       </div>
 
-      <template id="template-control" style="display: none;">
+      <template id="template-control" class="template-hidden">
```
gedankenstuecke, line 158: I only today learned about the `template` tag existing, super cool!

server/apps/main/templates/main/moderate_experience.html
```diff
@@ -306,13 +306,49 @@ <h5>Moderation reasons to be passed back to the author</h5>
...
+    <p>Please set the <i>Moderation status</i> to <i>in review</i> to hold reasons for later use, or set the status to <i>accepted</i> or <i>rejected</i> to send the reasons to the author now.</p>
+  </template>
+
+  <template id="template-text-reject-no-comments" class="template-hidden">
+    <p>You are rejecting the story without providing a reason. Are you sure you want to continue?</p>
+  </template>
+
+  <template id="template-text-approve-comments" class="template-hidden">
+    <p>You have included rejection moderation reasons, but are still aproving the story. Are you sure you want to continue?</p>
+  </template>
```
gedankenstuecke, line 329: I guess this is more a question of what we want: Should users be able to approve w/ rejection reasons given? 

I guess we also allow flagging triggering sections for which we apply the trigger label, right? In that case approving while labeling parts is expected behaviour?

server/apps/main/views.py
```diff
@@ -523,82 +524,113 @@ def moderate_experience(request, uuid):
         model = PublicExperience.objects.get(experience_id=uuid)
         if request.method == "POST":
             # get the data from the model
-            moderated_form = ModerateExperienceForm(request.POST)
-            moderated_form.is_valid()
-
-            # Get the (immutable) experience data
-            unchanged_experience_details = extract_experience_details(model)
```
gedankenstuecke, line 530: I know that this isn't where the `model` is introduced, but just to raise that it probably should be called `public_experience` or something similar as it's confusing to have such an abstract variable name :D

server/apps/main/views.py
```diff
@@ -523,82 +524,113 @@ def moderate_experience(request, uuid):
...
+                        change_reply=moderation_reply,
+                    )
+                except Exception:
+                    # if user writing E has deauthorized autspaces, delete public experience
+                    delete_PE(uuid=uuid, ohmember=user_OH_member)
+
+                # send a message to the author if they've requested it
+                moderation_status = data.get("moderation_status", "")
+                moderation_string = moderation_status.title()
+                try:
```
gedankenstuecke, line 565: I'm not 100% sure why we have to wrap this in `try/expect`, is it in case the `UserProfile` doesn't exist? Just trying to better understand the edge cases we're trying to catch here!

gedankenstuecke (commented):
Really well done @llewelld! I tried breaking it a bit with disabling my JS to bypass the client side form validation and even that seems to properly catch those edge cases as well, really impressive! 🕵️

I just left a few comments in-line where I wasn't 100% sure I understood what/why was happening. I guess most don't actually need code changes as much as just clarifications in the comments! 

But again, super impressive work! 🎉

server/apps/main/helpers.py
llewelld: This intention here was more to do with moderator workflow, building on the issue flagged by @mastoffel that the moderator can't 'edit' previous feedback. This change ensures that the form is pre-populated with the moderation replies from the _previous_ moderation round. This seems relevant especially for the 'in review' state, where the moderator may be returning to a half-completed review, but potentially also useful in the other cases as well.

But this is a change in behaviour, so up for discussion of course!

I had to check about the case you describe, where the user is bounced back after a failed validation, but in that case django seems to retain the old values automatically (so this change could be removed without affecting that).

server/apps/main/templates/main/moderate_experience.html
llewelld: AutSPACEs taught me about it too 😃

server/apps/main/templates/main/moderate_experience.html
llewelld: Yes, that was my thinking too. It could potentially make sense to flag issues, even for an approved story, in case the moderator is forcing a change to the triggering labels.

We could add a condition on this (check whether the labels have changed), or even just skip the check/dialogue in the case of an approval?

server/apps/main/views.py
llewelld: There are two possible exceptions that I'm aware of potentially being caught here. One of them, as you mention, is a potentially missing `UserProfile`. The other is in case reading of the `mod_message.txt` file fails for some reason, in which case an exception may bubble up from `get_moderator_message()`.

We could catch them separately to make this more explicit (or remove the `UserProfile` catch completely?) if you prefer?

server/apps/main/views.py
llewelld: Agreed; I've taken the opportunity to rename it! Thanks 😃

server/apps/main/helpers.py
gedankenstuecke: Ah of course, that makes total sense. I had forgotten about needing to re-populate the fields with existing bits! 👍

server/apps/main/templates/main/moderate_experience.html
gedankenstuecke: It might be a bit overkill to check for it through the JS client-side validation, but maybe what we should validate on the server side is that _approved_ moderated experiences can only contain moderation feedback that's trigger warning-related and not _ground for rejection_-related?

server/apps/main/views.py
gedankenstuecke: Given that the lack of `UserProfile` also creates the bug in #552, we maybe should make a helper to check if the user has a profile and that returns `None` (or similar) otherwise to help us check those things in conditionals without also having to manually check the presence of the `UserProfile`? As I think for this one we will have to keep checking its existence to make sure we can send a message. 

For the `mod_message.txt`, I guess given that our _default_ is part of the repo we could assume that it exists? :)

gedankenstuecke (commented):
Thanks for the replies, super useful! I left a few suggestions but those could be additional improvements to be handled at a later point!

server/apps/main/templates/main/moderate_experience.html
llewelld: Let me give this a go. Technically I think it'd be easier to catch "red+approve" submissions on the client side, because the moderation types come from the HTML template and the code doesn't actually know anything about them. It wouldn't be sensible to split the reasons across the code and templates, but we could send the type (red/amber) back as part of the POST submit. I'd also prefer not to have the check done just on the client side.

gedankenstuecke: Yeah, I wonder if we should move the checking into the DB backend somewhere which we could then also load on the client side? That way we could validate on both depending on the need. But that's probably it's own PR as it would be quite a big change! 

So I'd be happy to have a minimal adjustment here and make an improvement issue instead of squeezing it in here 🙂

server/apps/main/views.py
gedankenstuecke: Yeah, that makes sense to me!

gedankenstuecke (approved):

server/apps/main/templates/main/moderate_experience.html
llewelld: I've updated it so it only allows amber replies through on approval. There's a check at both ends (client submit and server form validation); it's not an ideal approach, but avoids more significant changes.

server/apps/main/views.py
llewelld: There's now a helper function for safely getting the user profile. I've used it here in the moderation view, but the #552 task is still needed for the other cases.

server/apps/main/templates/main/moderate_experience.html
gedankenstuecke: I only tried the check through JS in the frontend but that works really well. And the code for the backend looks great too!

server/apps/users/helpers.py
```diff
@@ -38,3 +38,18 @@ def user_submitted_profile(user):
...
+
+    Args:
+        user: request.user
+
+    Returns:
+        The UserProfile associated with the user if it exists, None otherwise
+    """
+    try:
+        uo = UserProfile.objects.get(user=user)
+    except:
```
gedankenstuecke, line 53: Minor comment, could we change this to be a more specific exception than the blanket `except`? Not too relevant as the only other thing than `DoesNotExist` would be if it returned >1 item, but still nicer to keep with the Python recommendations! 🙂

server/apps/main/views.py
gedankenstuecke: Perfect!

gedankenstuecke (approved):
Happy to have this merged, just left 2-3 minor suggestions but don't think we need to re-review for those!

server/apps/main/templates/main/moderate_experience.html
```diff
@@ -306,13 +306,49 @@ <h5>Moderation reasons to be passed back to the author</h5>
...
+
+  <template id="template-text-reject-no-comments" class="template-hidden">
+    <p>You are rejecting the story without providing a reason. Are you sure you want to continue?</p>
+  </template>
+
+  <template id="template-text-approve-comments" class="template-hidden">
+    <p>When approving a story you can cite <span class="word-amber">Amber</span> reasons, but not <span class="word-red">Red</span> reasons. Stories containing <span class="word-red">Red</span> reasons must be rejected.</p>
+  </template>
+
```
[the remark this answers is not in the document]
llewelld: I agree, this would be a sensible change. Just to avoid this PR losing focus I've created #553 to cover this as you suggested. It should be a straightforward change I think.

server/apps/users/helpers.py
llewelld: Yes, good point. I've refined it to just catch the `DoesNotExist` exception. We could also include the [`MultipleObjectsReturned`](https://docs.djangoproject.com/en/4.2/ref/models/class/#django.db.models.Model.MultipleObjectsReturned) case but I can't envisage a situation where this should happen for any good reason, so think it's best to leave it out so we know to look into it if it hapens. Does that seem okay to you?

gedankenstuecke: Absolutely, that's very sensible! 👍

gedankenstuecke:
From my side this looks good to go @llewelld!

llewelld:
Thanks as always for your careful review @gedankenstuecke.

gedankenstuecke:
Any time, hopefully it wasn't too nitpicky! 🤞

PR #551, opened 2023-07-13
upgrade packages

### LINKED ISSUE #548: Upgrade to django 4.2.3 or newer
### Summary

GitHub's [dependabot](https://github.com/advisories/GHSA-jh3w-4vvf-mjgr) informs that:

>  Django has regular expression denial of service vulnerability in EmailValidator/URLValidator 

I'm pretty sure we're not using either of these (and grep says the same), but we probably shouldn't be running a version with high severity security alerts anyway.

We're currently already on 4.2, so it should be an easy upgrade.

### What needs to be done?

- [ ] Update the dependences to django >= 4.2.3
- [ ] Check the docker deployment, unit tests and site still work after the upgrade.

### Who can help?

* Anyone with an interest in Poetry (the [python kind](https://python-poetry.org/), not the [other kind](https://peps.python.org/pep-0020/)).

---

gedankenstuecke:
Closes #548. 

@llewelld This was generated by just running `poetry lock` from the root directory (but requires having pyenv + poetry set up locally to match versions etc).

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -216,14 +216,14 @@ files = [
 
 [[package]]
 name = "certifi"
-version = "2022.12.7"
+version = "2023.5.7"
 description = "Python package for providing Mozilla's CA Bundle."
 category = "main"
 optional = false
 python-versions = ">=3.6"
 files = [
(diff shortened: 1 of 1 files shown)


gedankenstuecke:
In addition to the testing still passing: My cursory local testing seems to have worked as well

PR #554, opened 2023-07-19
Added authorship to model and form

### LINKED ISSUE #544: Update `PublicExperience` to include an "author" category
### Summary
When sharing an experience it will be necessary to know if the story about a user's experience or are they sharing it on behalf of someone else
### Deadline 
August 2023

### What needs to be done?
- [x] Decide on categories to be chosen from
- [x] Add this to the form and model
- [x] Update the share experience page
- [x] Update the view experiences page from the placeholder "AutSPACEs Contributor"

### Who can help?

Anyone interested

---

obinnaofomah:
Closes #544 Options for the authorship and relationship story


obinnaofomah:
@gedankenstuecke First attempt at authorship, and would appreciate feedback.

server/apps/main/forms.py
```diff
@@ -50,6 +50,16 @@ class ShareExperienceForm(forms.Form):
...
+    # authorship option
+    authorship = [
+        ("AutSPACEs contributor", "AutSPACEs contributor"),
+        ("Autistic individual", "Autistic individual"),
+        ("Parent of an autistic individual", "Parent of an autistic individual"),
+        ("Carer for an autisitc individual", "Carer for an autisitc individual"),
+    ]
+    experience_authorship = forms.ChoiceField(choices = authorship, widget = forms.Select(), required=False)
+    experience_authorship.group = 4
+
```
gedankenstuecke, line 62: I might be forgetting a discussion we had later, but I thought we had planned to go with a binary "Is this an experience shared on behalf of someone else?" and then Yes/No? 

And then having a second expansion, a free-text box which would record how/why the experience was submitted on behalf of someone else? 

At least that's the state we also have recorded in #544

helendduncan: Yes we did look at that when starting and then kinda glossed over it!
Will look into implementation now

gedankenstuecke: Okay, no worries! I really wasn't sure what the latest outcome of our discussion was! 👍

server/apps/main/forms.py
```diff
@@ -50,6 +50,16 @@ class ShareExperienceForm(forms.Form):
     research = forms.BooleanField(label = "Share for research", required=False)
     research.group = 3
 
+    # authorship option
+    authorship = [
```
helendduncan, line 54: So here something like 
`authorship = [("Someone else", "Yes"), ("Myself", "No")]`

 `authorship_relation = forms.ChoiceField(widget=forms.RadioSelect, choices=authorship, label="")`

could work - there will be two options to the question _Is this an experience shared on behalf of someone else?_ which would be saved as "Someone else" or "Myself" in the form data. 

Then have a free-text box that is disabled unless "Someone else" is clicked - could do a javascript _onchange_ to manage that

llewelld:
For the appearing and disappearing text entry field that we discussed during our co-working session, I had to do something similar for the user profile page. Here are some links in case it's possible to generalise this for both cases. This isn't yet merged in, but is forming part of #556.

1. The javascript file that's imported to do the work: https://github.com/llewelld/AutSPACEs/blob/iss543_userprofile_changes/static/js/activate-profile.js
2. The page it gets loaded into: https://github.com/llewelld/AutSPACEs/blob/iss543_userprofile_changes/server/apps/users/templates/users/profile.html#L11

gedankenstuecke:
I think this PR can now be closed as I think the authorship work is already merged from a different branch? 🙂

helendduncan:
agreed - this can be closed

PR #555, opened 2023-07-19
Use get_user_profile to get the UserProfile object

### LINKED ISSUE #552: [BUG]: Viewing public experience crashes for Anon. users
### Describe the bug

Currently it is not possible to view the publicly shared experiences unless one is logged in. As a logged-out user one encounters the following error:

`Field 'id' expected a number but got <SimpleLazyObject: <django.contrib.auth.models.AnonymousUser object at 0x400f471df0>>.`

The problem is that in line 335 of the `views.py` we load the user profile as `profile = UserProfile.objects.get(user=request.user)`, but non-logged in users have no expanded profile, so we need to catch that case :-) 

### Steps to reproduce

1. Log out of AutSPACEs, 
2. browse to `/main/public_experiences`
3. et voilà 😉 

### In which environment are you using AutSPACEs?

_No response_

### Other comments

_No response_

llewelld:
The get_user_profile helper function will return None of the user profile doesn't exist, rather than triggering an exception. These changes switch to using the helper function to avoid potentially errors iif the user isn't logged in.

Also adds a unit test to catch the error case.

Fixes #552.

### DIFF
--- a/server/apps/main/tests/test_views.py
+++ b/server/apps/main/tests/test_views.py
@@ -532,6 +532,8 @@ class Views(TestCase):
         c = Client()
         c.force_login(self.user_a)
         response = c.post("/main/public_experiences/")
+        assert response.status_code == 200
+        self.assertTemplateUsed(response, "main/experiences_page.html")
 
         # Check that there is only one story visible - the one with no triggering labels
         for item in response.context[0]:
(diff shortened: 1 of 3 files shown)


gedankenstuecke (approved):
That looks great, @llewelld! And thanks for also expanding the tests to include covering that case!

obinnaofomah (approved):

Issue #543, opened 2023-07-06
User Profile changes 


helendduncan:
### Summary
Following discussions the July 2023 meet up ([Google doc](https://docs.google.com/document/d/1mYcTs0kfh8HWK1X87Wc5whARrhH_Rm9xaw2a-uT6UFs/edit#heading=h.7f28vh3w1oai)). There have been some suggestions to the current set up of the user profile. 
This issue is an attempt to capture the feedback and form actionable issues - however it may be better suited to the [discussions](https://github.com/alan-turing-institute/AutSPACEs/discussions) section

### Deadline 
September 2023

### What needs to be done?
With the caveat that these are subject to modification as we try to interpret the issue fully
#### Overview
- [ ] Provide context for asking for specific demographic data - Is it for research? Platform development? Something else?

#### Technical
- [ ] Set the question regarding autism status to be mandatory

#### Additional demographic data
 - [ ] Ethnicity for research purposes
 - [ ] Socio-economic background

#### Questions
- [ ] Is there the ability to change your answer to the user profile questions
- [ ] How do you treat profiles who "prefer not to say" for the autism status
- [ ] 

### Who can help?

* 
* 

---

gedankenstuecke:
Thanks for this great write-up @helendduncan! I'll take a stab at writing up my preferred/suggested approach to the demographics-part of the profile by listing the questions/labels etc., as I think that this might be the easiest approach:

### Item 1: Do you identify as autistic? (mandatory field)

**Sub-label/explanation text**: AutSPACEs is focusing on collecting and sharing the voices and lived experiences of autistic people, which is why this is a mandatory question. If you select _prefer not to answer_ your answers will be considered as coming from a non-autistic person, labeled as such and e.g. not used for research

**Form type**: Dropdown menu, with answer options _Yes_, _No_, _Prefer not to say_

### Item 2: What is your age group? 

**Sub-label/explanation text**: Providing your age will help researchers understand potential trends across age groups. It will also help readers to better understand your perspective if this data is made public.

**Form type**: Dropdown menu, with existing age categories, __except__ removing the <18 group, as we don't want to deal with the added issues of data from minors. Additionally, provide a toggle switch for _make public: yes/no_

### Item 3: What gender do you identify with?

**Sub-label/explanation text**: Providing your gender identity will help researchers understand potential trends across this demographic. It will also help readers to better understand your perspective if this data is made public.

[...]

PR #556, opened 2023-07-25
Update user profile fields

llewelld:
Updates the fields shown on the user profile page following the July 2023 meetup.

Contributes to issue #543.


llewelld:
This should be in a reasonable state now, so I'm removing it from draft. I tried to cover all the elements described in [the issue](https://github.com/alan-turing-institute/AutSPACEs/issues/543#issuecomment-1625178816). The one part I've not tackled is the following:

> Additional questions: During the meeting we discussed whether to have a question on "(additional) relationship to autism", e.g. being a parent/carer/friend/co-worker etc. of someone with autism? I think generally it would be good to have this, maybe could fit in between Items 1 & 2?

I skipped this because I'd need a bit more direction on it: ideally the specific items to add to the drop down list, whether a free-text field is needed for any of the cases and whether there should be a 'make public' toggle. It looks like it also potentially relates to PR #554 that @obinnaofomah and @helendduncan are working on.

I'm very happy to make these changes, but if it makes it easier I could look at this in a separate PR.

server/settings/environments/development.py
```diff
@@ -73,15 +73,35 @@ def _custom_show_toolbar(request):
...
+    'netdna.bootstrapcdn.com',
+    'cdn.jsdelivr.net',
+)
+CSP_STYLE_SRC = (
+    "'self'",
+    "'unsafe-inline'",
+    'fonts.googleapis.com',
+    'maxcdn.bootstrapcdn.com',
+    'netdna.bootstrapcdn.com',
+    'cdn.jsdelivr.net',
```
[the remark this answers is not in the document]
gedankenstuecke: I think that's fine, I guess we should at some point also check the production settings for that (hoping that this is a mental note enough to remember when things break during deployment 😂)

server/apps/main/templatetags/custom_tags.py
```diff
@@ -32,3 +32,8 @@ def url_replace(request, field, value):
     query = request.GET.copy()
     query[field] = value
     return '?{}'.format(query.urlencode())
+
+@register.filter
+def public_toggle(items, index):
```
gedankenstuecke, line 37: I'm not 100% sure I understand what's going on here. What exactly does the public_toggle do/check for? :)

server/apps/users/forms.py
```diff
@@ -12,47 +24,57 @@ class UserProfileForm(forms.Form):
...
-    autistic_identification.group = 1
 
-    description = forms.CharField(label="Description", max_length=2048, strip=True, required=False,
-                                 widget=forms.Textarea(attrs={"placeholder":"An optional description of yourself",
+    gender_self_identification = forms.CharField(label="Self identification",
+                                 max_length=150, strip=True, required=False,
+                                 widget=forms.TextInput())
+
+    gender_self_identification.group = 1
+    gender_self_identification.layout_horizontal = True
```
gedankenstuecke, line 57: The JS hide/show of the free-text field for self-identification works. But what doesn't seem to work (at least for me in Firefox) is actually filling in the text. The JS console doesn't give any error or anything, but I can't seem to actually type into the box 🙈

gedankenstuecke (commented):
Thanks so much for this @llewelld, it looks really nice and I like the touch of the hover-info boxes for explaining why we collect that data. My only concern would be how well those work on mobile/how accessible they are, but we can test that later and simplify if needed! 

>>Additional questions: During the meeting we discussed whether to have a question on "(additional) relationship to autism", e.g. being a parent/carer/friend/co-worker etc. of someone with autism? I think generally it would be good to have this, maybe could fit in between Items 1 & 2?
>
>I skipped this because I'd need a bit more direction on it: ideally the specific items to add to the drop down list, whether a free-text field is needed for any of the cases and whether there should be a 'make public' toggle. It looks like it also potentially relates to PR https://github.com/alan-turing-institute/AutSPACEs/pull/554 that obinnaofomah and helendduncan are working on.

Yeah, I think it's fine to not incude these aspects in this inital PR, same for adding the display of public labels in the front-end of experiences! 

Otherwise, see inline comments for a couple of questions! 🙂

server/apps/users/forms.py
llewelld: Ooops, thanks for checking. An error in refactoring meant the disabled flag got set in reverse. It should be fixed now.

server/apps/main/templatetags/custom_tags.py
llewelld: This was used to condition on the 'Make public' fields. But it was hacky and I should have put more effort into coming up with a bettter alternative. I've made a more general filter for this now so all the toggle fields can be handled the same way on the form.

gedankenstuecke: Ah, so this is the bit which in this PR replaces the regular checkbox with the toggle switch, right? :)

gedankenstuecke (approved):
Looks and works great now! 👍

server/apps/main/templatetags/custom_tags.py
llewelld: Yes; pretty much (the actual work is done in the template &mdash; which I know you know 😄 &mdash; but this helps simplify the conditioning).

Issue #536, opened 2023-07-04
[FEATURE]: Add function for account deletion


gedankenstuecke:
### Feature for…

Both

### What is the problem you want to solve?

Currently users can not fully delete their account on AutSPACEs. The best workaround users currently would have are the following: 

1. Delete all of their (public) experiences / set them to private to avoid them being stored in the AutSPACEs DB
2. Empty the `UserProfile` information manually by resetting to N/A
3. Go to _Open Humans_ and de-authorize AutSPACEs, thus removing AutSPACEs' permission to read/write/message through OH

Even doing all these this would leave some (if useless) data back in AutSPACEs. 

### How do you want to solve the problem?

It would be better if we allowed users to delete their AutSPACEs account alongside all data from the AutSPACEs frontend. I propose doing so on the `UserProfile` page, which already is labeled _Account_ page. To do so we would at least need to: 

1. Add a new route in the `urls.py` for a `delete_account` view 
2. Add a button to the _UserProfile_ page that does a POST/DELETE request on this new view

Ideally this solution should also offer the following: 
1. Request confirmation from the user as it's a very destructive action
2. Before the actual deletion ask the user whether they would like us to also delete their AutSPACEs data on Open Humans

PR #557, opened 2023-07-26
Add an account deletion option

llewelld:
Adds an option to delete the user's account to the profile page.

### DIFF
--- a/server/apps/users/forms.py
+++ b/server/apps/users/forms.py
@@ -114,3 +114,7 @@ class UserProfileForm(forms.Form):
         widget=forms.CheckboxInput(attrs={"class": "custom-control-input"}))
     profile_submitted.group = "hidden"
 
+class UserProfileDeleteForm(forms.Form):
+    delete_oh_data = forms.BooleanField(label = "Delete your stories from OpenHumans",
+                                       required=False,
+                                       widget=forms.CheckboxInput(attrs={"class": "custom-control-input"}))
--- a/server/apps/users/helpers.py
+++ b/server/apps/users/helpers.py
@@ -1,5 +1,6 @@
 from django.contrib.auth.models import User
 from .models import UserProfile
+from server.apps.main.models import PublicExperience
 
 def user_profile_exists(user):
     """
@@ -53,3 +54,23 @@ def get_user_profile(user):
     except UserProfile.DoesNotExist:
         uo = None
     return uo
+
+def delete_user(user, delete_oh_data):
+    """
+    Deletes the user and all data associated with it.
+
+    Args:
+        delete_oh_data: True if stories on OpenHumans should also be deleted
+
+    Returns:
+        None
+    """
+    ohmember = user.openhumansmember
+
+    # Delete the stories from the OpenHumans database
+    if delete_oh_data:
+        ohmember.delete_all_files()
+
+    # Delete the actual user
+    user.delete()
+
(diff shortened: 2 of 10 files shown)


llewelld:
I'm thinking it might make sense to create a dummy form for account deletion in order to benefit from django's CSRF protection.

gedankenstuecke:
Great point re: the form and CSRF! Just wanted to add to this, saying that we don't have to make a "proper" Django form for this to work. We can just make a minimal form in the HTML template that consists of the CSRF token and and the button and it should work too! (c.f. https://docs.djangoproject.com/en/4.2/howto/csrf/). 

Sorry if this is super obvious and was what you had planned for in any case, but the "dummy form" made me think of the Django `Form` class!

llewelld:
> Sorry if this is super obvious and was what you had planned for in any case, but the "dummy form" made me think of the Django `Form` class!

I was just reading up the docs on it, so this is useful info to know.

llewelld:
This should now be ready for review, so I'm removing the draft status.

llewelld:
Fixes #536.

server/apps/users/forms.py
```diff
@@ -92,3 +92,7 @@ class UserProfileForm(forms.Form):
         widget=forms.CheckboxInput(attrs={"class": "custom-control-input"}))
     profile_submitted.group = "hidden"
 
+class UserProfileDeleteForm(forms.Form):
+    delete_oh_data = forms.BooleanField(label = "Delete your stories from OpenHumans",
+                                       required=False,
+                                       widget=forms.CheckboxInput(attrs={"class": "custom-control-input"}))
```
gedankenstuecke, line 98: Oh, I'd have totally forgotten about that, great catch!

server/apps/users/helpers.py
```diff
@@ -53,3 +54,35 @@ def get_user_profile(user):
...
+
+    # Delete the user's profile
+    UserProfile.objects.filter(
+        user=user
+    ).delete()
+
+    # Delete the actual user
+    User.objects.filter(
+        id=user.id
+    ).delete()
```
gedankenstuecke, line 87: I think all of these except for the `oh_member.delete_all_files()` are not necessary actually, as I'm 99% sure that all of our models do `on_delete=models.CASCADE`, which means that a deletion of the main `User` will also delete all the other objects!

gedankenstuecke (commented):
Great work as usual @llewelld! I left one comment where I think that the helper could potentially be simplified quite a lot!

gedankenstuecke:
Oh, and I gotta admit I didn't run it to test it yet, as I was already out of office when I got the email. So let's pick this up next week ☺️ 🏖️

server/apps/users/helpers.py
llewelld: Great; and the unit tests seem to confirm it. I've pushed a simplifed version.

server/apps/users/helpers.py
```diff
@@ -53,3 +54,25 @@ def get_user_profile(user):
...
+    ohmember = user.openhumansmember
+
+    # Delete the stories from the OpenHumans database
+    if delete_oh_data:
+        ohmember.delete_all_files()
+
+    # Delete the actual user
+    User.objects.filter(
+        id=user.id
+    ).delete()
```
gedankenstuecke, line 77:
```suggestion
    user.delete()
```

The existing method of course works too, but one can also delete instances of the class/model by just calling delete on the instance, instead of making another DB query 🙂

gedankenstuecke (approved):
Looks great, happy to have it merged in as is, or with the minor modification I suggested 👍

llewelld:
Rebased.

Issue #535, opened 2023-07-04
Support users to submit their first experience


GeorgiaHCA:
### Summary

There is evidence from previous focus groups that the research question is quite broad and it may be unclear or ambiguous to new users what to enter and how to enter it. It would be helpful to have a walkthrough, tutorial or other method to support people to enter stories - create an AutSPACEs onboarding process. 

### What needs to be done?

- [ ] Look at existing citizen science projects and how they onboard people
- [ ] Research into what is possible
- [x] Create a diagram or diagrams to support onboarding

### Who can help?
 
* Please  add any suggestions or feedback in the comments below. 

---

gedankenstuecke:
Here's a few JS walkthrough libraries: 

- [IntroJS](https://introjs.com/)
- [ShepherdJS](https://shepherdjs.dev/)
- [TourGuide](https://tourguidejs.com/)

Having said that, I don't have 1st-hand experience with either of them!

GeorgiaHCA:
I've made a basic flow chart diagram using [draw.io](https://drive.google.com/file/d/16nX5IvyiaQ9DgzJlnE3mfnxHmWxLsk2o/view?usp=drive_link) - would it be helpful to include screenshots?

gedankenstuecke:
Do you mean including screenshots of the flowcharts here? Or including screenshots in the flowcharts? 

Oh, and the draw.io link requires permission to view!

GeorgiaHCA:
I was meaning that it could be helpful to users to have screenshots of the platform alongside the steps in the flowchart, maybe with arrows or elements highlighted? Here's another link to the flowchart: https://drive.google.com/file/d/1J3zaYyY01WCvc0On6Q2LxAB7xvq2o4UU/view?usp=sharing

gedankenstuecke:
thanks, the link now works! I wouldn't add screenshots into the flowchart, as I think the point would be to have a quicker overview of the process and adding screenshots could expand on that too much?

GeorgiaHCA:
Hi @llewelld  - I think a couple of sentences would be enough for now in the Help section. Maybe something like: 

To share an experience on AutSPACEs, please go to the share experience page when you are logged in and write about any time you have been effected by your senses in the text box. Then you can submit your experience for research or to share publicly on our website. The flow chart below shows the steps you can take to share an experience on AutSPACEs. 

How does that sound @llewelld, @gedankenstuecke ?

gedankenstuecke:
[...]

PR #561, opened 2023-08-02
Add Story Sharing help

llewelld:
Adds a flow chart to illustrate the story sharing process, along with some explanatory text, to the help pages.

Contributes to #535.


llewelld:
I've added the drawio source to the `static/images` folder, but I'm certain this is the wrong place, (we don't need it served on the final site). Any suggestions for where it should go? Maybe we don't need it stored in the repo at all?

llewelld:
A preview:

image

gedankenstuecke (approved):
That's a great addition! 

Regarding the question of where to put the source file: It's an excellent question. Generally I'd say somewhere in the `00-project-documentation` folder could be a good fit. It's a bit hidden but `00-project-documentation/platform-design/platform-design/platform-content` seems like it could fit?

gedankenstuecke (approved):
Thanks, then I'd say this is good to go!

Issue #540, opened 2023-07-06
Return to previous page from Single Story page


helendduncan:
### Summary

Following the July 2023 meet up ([Google doc](https://docs.google.com/document/d/1mYcTs0kfh8HWK1X87Wc5whARrhH_Rm9xaw2a-uT6UFs/edit#heading=h.7f28vh3w1oai)): We received feedback that it would be useful if a "back to previous page in same state" button was implemented in the single stories page

### Deadline 
September 2023

### What needs to be done?

- [ ] Implement button
- [ ] Decide on behaviour if story directly linked (go straight to public experiences page?)

### Who can help?

Anyone is welcome to, might clash a bit with work on #531 so mild suggestion of @helendduncan 

---

gedankenstuecke:
I'd say linking back to the public experiences page sounds reasonable to me for the cases where people were opening it from the link. We could also rename the button label to "view all experiences" or something similar in those cases?

llewelld:
Could this be implemented using a `history.back()`? That would hopefully avoid the need to store complex state (django would handle this automatically?) while also ensuring the user's history order is mantained.

https://www.w3schools.com/jsref/met_his_back.asp

gedankenstuecke:
`history.back()` would work for users that come from inside AutSPACEs, but if you are deep-linked to an experience page it would either take you back to wherever else or nowhere in case it's a new window/tab. 

I wonder if we should start using the Django sessions to store all those search/filter parameters: https://docs.djangoproject.com/en/4.2/topics/http/sessions/ 

That way we could restore the states more easily. (Coming to think of it: This would also allow us to avoid having to pass the titles of experiences in URLs as we have for the experience-deletion confirmation!)

gedankenstuecke:
I've implemented the latter (& unrelated) part of this (the passing of the titles in URLs) in #568.

I think the same approach could be taken to store information for the transition between the full _show experiences_ view, the individual experience view and going back. 

For this to work we'd have to store two sets of parameters in the `request.session`, the `trigger_opt_ins` and the `page`, which would allow us to recreate the full request combination of page & labels. 

These session parameters could be set/updated every time a user visits the `show_experiences` page. In case a user is logged in and doesn't have any trigger labels set just yet, we could load them from the `UserProfile` and otherwise start from the blank slate. 

Does that sound reasonable?

helendduncan:
That sounds really sensible

PR #568, opened 2023-08-09
Use sessions for storing experience title

gedankenstuecke:
This PR addresses a `TODO` item that was left in the code: 

Trying to delete an experience takes a user from the list view to a new confirmation page. On that page we want to display the title of the experience without having to query Open Humans again. So far this is achieved by passing the experience title as a URL parameter: 

`path("delete/<uuid>/<title>/", views.delete_experience, name="delete_exp"),`

This is a bit clunky and could potentially break depending on the characters people use in their titles etc (though we did some proper character escaping). 

A better solution is to store these titles somewhere locally in somewhat ephemeral storage. Django provides a `session` backend for such purposes and we already use that middleware implicitly without actually making use of it. In our current setup the Postgres database has a session store. I've updated the `my_stories` and `delete_experience` views to make use of this: 

- When loading the `my_stories` view a helper now sets a session dictionary that links all experience `uuid`s to the experience titles. 
- The `delete_confirmation` view we can than grab this data from the session without having to pass anything else around, both on `GET` and `POST`

@helendduncan I'm flagging you as I think this relates – by a potential similar approach – to the question of how to handle people going back from an individual experience page to the main listing and how/where to store all the trigger warning opt-ins/outs (c.f. #531 and #540)

### DIFF
--- a/server/apps/main/views.py
+++ b/server/apps/main/views.py
@@ -289,8 +289,7 @@ def delete_experience(request, uuid):
     """
     Delete experience from PE databacse and OH
     """
-    # TODO: we currently are passing title via url because it is nice to display it in the confirmation. We could improve the deletion process by having a javascript layover.
-
+    
     titles = request.session.get('titles', {})
     title = titles.get(uuid, "no title")
 
--- a/server/apps/main/views.py
+++ b/server/apps/main/views.py
@@ -48,6 +48,7 @@ from .helpers import (
     number_stories,
     get_message,
     message_wrap,
+    experience_titles_for_session,
 )
 
 from server.apps.users.helpers import (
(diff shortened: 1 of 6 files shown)


server/apps/main/views.py
```diff
@@ -284,12 +285,15 @@ def view_experience(request, uuid):
 
 
 # @vcr.use_cassette("server/apps/main/tests/fixtures/delete_exp.yaml", filter_query_parameters=['access_token'])
-def delete_experience(request, uuid, title):
+def delete_experience(request, uuid):
     """
     Delete experience from PE databacse and OH
     """
     # TODO: we currently are passing title via url because it is nice to display it in the confirmation. We could improve the deletion process by having a javascript layover.
```
helendduncan, line 292: My only comment would be that it looks like this TODO can now be removed?

PR #578, opened 2023-08-30
Remind the user to complete their profile

### LINKED ISSUE #534: Remind the user to fill out their profile
### Summary

Figure out a sensible flow in case the user never fills out their profile.

There may be multiple reasons why a user unintentionally doesn't fill out profile information:
1. In case the functionality that redirects the user to the Profile page on first entry fails, the user may be unaware that they have to.
2. The user may want to get straight to the content, skip profile creation, and later forget about it.

We should decide on an appropriate flow to try to avoid this situation arising. For example, if the `UserProfile` is found not to exist, or if the user has never submitted a profile, it might make sense to redirect them to the profile page at various points. The points at which it makes sense to do this will need to be decided (e.g. after a period of time; when the user next logs in; when the user performs specific actions; etc.)

### What needs to be done?

- [ ] Decide on when appropriate triggers for reminding the user.
- [ ] Implement them.

### Who can help?

* Wider community for high-level design
* Developers for technical implementation.

---

llewelld:
If the user submits an experience with either the research or public flags set, but hasn't yet completed their user profile, this change will redirect the user to the profile page with a suitable message requesting that they complete it.

Fixes #534.

### DIFF
--- a/server/apps/main/tests/test_views.py
+++ b/server/apps/main/tests/test_views.py
@@ -310,6 +310,127 @@ class Views(TestCase):
         filter_query_parameters=["access_token"],
         match_on=["path"],
     )
+    def test_share_exp_redirect_confirmation(self):
+        """
+        Test that the user is redirected to the user profile
+        given the profile is incomplete and the experience will be
+        shared publicly and for research
+
+        logic:
+            request.user.is_authenticated == True
+            request.method == "POST"
+            form.is_valid() == True
+            viewable == True
+            research == True
+            user.autistic_identification == ''
+        """
+        c = Client()
+        c.force_login(self.user_a)
+
+        user_a_stories_before = len(
+            PublicExperience.objects.filter(open_humans_member=self.oh_a)
+        )
+
+        response = c.post(
+            "/main/share_exp/",
+            {
+                "experience_text": "Here is some experience text",
+                "difference_text": "Here is some difference text",
+                "title_text": "A new story added",
+                "viewable": "False",
+                "research": "False",
+                "open_humans_member": self.oh_a,
+            },
+            follow=True,
+        )
+
+        user_a_stories_after = len(
+            PublicExperience.objects.filter(open_humans_member=self.oh_a)
+        )
+
+        pe_db_after = PublicExperience.objects.filter(title_text="A new story added")
+
+        # Check that a story with the title now exists in the database
+        assert len(pe_db_after) == 0
+
+        # Check that there is one new story for user A
+        assert user_a_stories_after == user_a_stories_before
+
+        # Check that there is a redirect after
+        assert response.status_code == 200
+
+        # Check that the user is redirect to confirmation page and not
+        # the profile page
+        self.assertRedirects(response, "/main/confirm_page/",
+                             status_code=302, target_status_code=200)
+        self.assertTemplateUsed(response, 'main/confirmation_page.html')
+
+    @vcr.use_cassette(
+        "server/apps/main/tests/fixtures/share_experience.yaml",
+        record_mode="none",
+        filter_query_parameters=["access_token"],
+        match_on=["path"],
+    )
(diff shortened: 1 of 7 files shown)


llewelld:
This will need to be rejigged slightly after #569 has been merged to move the updated profile messages into the new profile messages file.

llewelld:
Switching to draft while I update the unit tests.

server/apps/main/tests/test_views.py
```diff
@@ -304,6 +304,127 @@ def test_share_exp_submit_new_experience(self):
...
+        )
+
+        user_a_stories_after = len(
+            PublicExperience.objects.filter(open_humans_member=self.oh_a)
+        )
+
+        pe_db_after = PublicExperience.objects.filter(title_text="A new story added")
+
+        # Check that a story with the title now exists in the database
+        assert len(pe_db_after) == 0
```
gedankenstuecke, line 354: Maybe I'm having a brainfart, but shouldn't this be `len==1` if the story exists?

server/apps/main/tests/test_views.py
```diff
@@ -304,6 +304,127 @@ def test_share_exp_submit_new_experience(self):
...
+
+        # Check that there is one new story for user A
+        assert user_a_stories_after == user_a_stories_before
```
gedankenstuecke, line 357: Same here, doesn't this validate that we have 0 experiences after adding one?

server/apps/main/tests/test_views.py
```diff
@@ -304,6 +304,127 @@ def test_share_exp_submit_new_experience(self):
...
+        "server/apps/main/tests/fixtures/share_experience.yaml",
+        record_mode="none",
+        filter_query_parameters=["access_token"],
+        match_on=["path"],
+    )
+    def test_share_exp_redirect_confirmation(self):
+        """
+        Test that the user is redirected to the user profile
+        given the profile is incomplete and the experience will be
+        shared publicly and for research
```
gedankenstuecke, line 317: Based on the function title (& logic below) I guess that this one actually tests that the redirect is to the confirmation page?

server/apps/main/tests/test_views.py
gedankenstuecke: Nevermind, I guess I understood now, It's 0 for both as the experience is not public, it's just that the comments need to be adjusted to reflect what we actually test for :)

gedankenstuecke (changes requested):
Overall this works really great, well done @llewelld! I left some in-line comments for the tests, where the doc-strings & comments don't match what we test for right now. 

Otherwise there's a slightly larger question about the approach for how to send the "message" to the user when the redirect to the profile happens. Currently you've made a partial for the profile messages and embedded that directly in the `profile.html` where it will be shown when appropriate. 

What I had in mind was using the actual `messages` feature of Django instead: 

In our main `application.html` we already have defined a place where those messages can be shown as boxes, c.f. https://github.com/alan-turing-institute/AutSPACEs/blob/1405bf11afbf7470be457fc0a22811b1b348516a/server/apps/main/templates/main/application.html#L36 

Currently this is only used to raise errors up as messages when trying to submit an experience, c.f. https://github.com/alan-turing-institute/AutSPACEs/blob/1405bf11afbf7470be457fc0a22811b1b348516a/server/apps/main/views.py#L186

While we might want to change this to the Bootstrap validation also used for the forms eventually for consistency, I wonder if using the messages approach might be a good alternative to avoid repeating the message interface code and being able to define it only once? 

Having said all that, I don't have really strong opinions either way, and there might be good reasons to implement it using the partials that i've overlooked!

llewelld:
After user testing and discussion, we're going to add in a 'Success' page after the user submits a Story, so this nudging will have to be moved to this success page in this case. Consequently it doesn't make sense to keep this so I'm closing it.

PR #579, opened 2023-08-31
Add configs for user test deployment to fly.io

gedankenstuecke:
This PR tweaks our config slightly and adds the files necessary to deploy a test environment to fly.io to allow participants in the user testing to explore the website without having to share remote control via zoom. 

It adds/changes: 

1. The `Dockerfile` used for the fly.io deployment 
2. The `fly.toml` for the config on their end
3. The `dj-database-url` package which loads the postgres database from the `DATABASE_URL` environment variable that fly uses
4. edits the `common.py` config to detect if a `DATABASE_URL` is set in order to identify that we're on fly.io to then use that config rather than the local dev setup.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,26 @@
+ARG PYTHON_VERSION=3.9.16-slim-buster
+
+FROM python:${PYTHON_VERSION}
+
+ENV PYTHONDONTWRITEBYTECODE 1
+ENV PYTHONUNBUFFERED 1
+
+# install psycopg2 dependencies.
+RUN apt-get update && apt-get install -y \
+    libpq-dev \
+    gcc \
+    && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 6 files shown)


llewelld (approved):
I didn't test this on fly.io, but it didn't cause any problems for the local docker run. Would you like me to run tests on fly.io just as well? Let me know if you would; otherwise consider it approved.

gedankenstuecke:
Thanks @llewelld, I think being able to still run it locally is good enough, given that the deployment to fly.io worked okay with the settings as in this PR! 

Having said that, if you're curious to see if fly.io really works that easy as advertised please feel free to test it out! I think with the settings here it actually should be quite straight-forward, with the exception being that one needs to manually set all the variables we have in the `.env` in the fly.io interface as otherwise the `fly deploy` step will crash due to missing settings :)

PR #580, opened 2023-08-31
Add authorship

### LINKED ISSUE #544: Update `PublicExperience` to include an "author" category

helendduncan:
Closes #544 

Adds authorship information to the model and the forms and displays authorship on stories

Bug: When viewing experience, the authorship relationship is not disabled - could this be due to js visibility?


server/apps/main/forms.py
```diff
@@ -63,6 +63,20 @@ class ShareExperienceForm(forms.Form):
     moderation_status = forms.ChoiceField(choices = statuses, widget = forms.Select(), required=False)
     moderation_status.group = "hidden"
 
+    authorship_choices = [
+        (True, "Experience is my own"),
+        (False, "Experience is someone else's"),
+    ]
+    first_hand_authorship = forms.ChoiceField(choices = authorship_choices, widget = forms.RadioSelect(), required=False)
```
gedankenstuecke, line 70: I think this should actually be `required=True`, as otherwise it will crash as it allows for submitting the form without a choice (which is `""`) and then crashes as the model doesn't validate that: 

```
ValidationError at /main/share_exp/

['“” value must be either True or False.']

 /code/server/apps/main/views.py, line 173, in share_experience 
```

server/apps/main/templates/main/share_experiences.html
```diff
@@ -179,28 +179,45 @@ <h3><label for="{{ field.id_for_label }}">{{ field.label }}</label></h3>
...
+            <div class="form-check {{ columns }}">
               {% if field.label == "Other"%}
                 <label for="{{ field.id_for_label }}">{{ field.label }}:</label>
                 {{field}}
+              {% elif field.label == "First hand authorship" %}
+              <label for="{{ field.id_for_label }}"></label>
+                {{ field }}
+              {% elif field.label == "Relationship"%}
+                <label for="{{ field.id_for_label}}">{{ field.label }}:</label>
+                {{ field }}
```
gedankenstuecke, line 212: I'm not sure I understand why we have to define these exceptions, they seem to look identical to the regular `{% else %}` option, no?

static/js/set_relationship_vis.js
```diff
@@ -0,0 +1,35 @@
...
+    });
+
+    // Set the initial visibility
+    // alert($('input[name="first_hand_authorship"]:checked').val()==undefined)
+    set_visibility($('input[name="first_hand_authorship"]:checked').val()!=undefined, '#id_authorship_relation', true)
+
+    // Control the visibility of a field
+    function set_visibility(visible, field_id, instant) {
+        duration = instant ? 0 : 200;
+        $(field_id).prop('disabled', !visible);
```
gedankenstuecke, line 26:
```suggestion
```

Removing this line seems to make the `disabled` property work properly for the read-only view?

server/apps/main/forms.py
```diff
@@ -63,6 +63,20 @@ class ShareExperienceForm(forms.Form):
...
+    first_hand_authorship.group = 4
+
+    authorship_relation = forms.CharField(label="Relationship",
+                                 max_length=300, strip=True, required=False,
+                                 widget=forms.TextInput(),
+                                 help_text='For example: "I am the non-autistic parent of an autistic adult", "I am the autistic parent of autistic children", "I am a teacher who works with autistic people", "I am best friends of an autistic person". - Please do not share personally identifying information.')
```
gedankenstuecke, line 76: This is not required on the form-side of things, but once someone picks "reporting on behalf of someone else" it should be mandatory i think. But that probably can't easily be verified inside the form setup here. :)

static/js/set_relationship_vis.js
gedankenstuecke: Actually, that change seems to break the correct invisibility setting  on load :/

gedankenstuecke (changes requested):

server/apps/main/forms.py
helendduncan: Actioned in [6cb302d](https://github.com/alan-turing-institute/AutSPACEs/pull/580/commits/6cb302d57d5b05def4bf729a7fb85487a9095e9c)

server/apps/main/templates/main/share_experiences.html
gedankenstuecke: Actually, re-viewing it now, is this because the label needs to come before the field in this case?

server/apps/main/forms.py
helendduncan: Could do something like [this](https://docs.djangoproject.com/en/4.2/ref/forms/validation/#cleaning-and-validating-fields-that-depend-on-each-other) so it would get caught in validation?

gedankenstuecke: yes, that looks good!

server/apps/main/forms.py
```diff
@@ -77,6 +77,21 @@ class ShareExperienceForm(forms.Form):
...
+        # Raise an error if first-hand authorship has a relationship description
+
+        cleaned_data = super().clean()
+
+        first_hand_authorship = cleaned_data.get("first_hand_authorship")
+        authorship_relation = cleaned_data.get("authorship_relation")
+
+        if first_hand_authorship == "False" and authorship_relation == "":
+                self.add_error("first_hand_authorship", ValidationError(("Stories written on behalf of someone else must have a description of the relationship"), code="invalid"))
+        if first_hand_authorship == "True" and authorship_relation != "":
```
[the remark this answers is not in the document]
helendduncan: From a brief look around the docs it might not be straightforward as it depends on two fields so those checks happen after the individual field form validation but I will give it a shot - I guess plan B would be a more verbose error message?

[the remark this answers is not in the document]
gedankenstuecke: Yeah, that's fair! Alternatively we could do that "removal" of the entered text using JS in the front-end on submission?

server/apps/main/forms.py
gedankenstuecke: Actually, I think this was accidentally reverted again somewhere!

server/apps/main/forms.py
```diff
@@ -63,6 +63,35 @@ class ShareExperienceForm(forms.Form):
     moderation_status = forms.ChoiceField(choices = statuses, widget = forms.Select(), required=False)
     moderation_status.group = "hidden"
 
+    authorship_choices = [
+        (True, "Experience is my own"),
+        (False, "Experience is someone else's"),
+    ]
+    first_hand_authorship = forms.ChoiceField(choices = authorship_choices, widget = forms.RadioSelect(), required=False)
```
gedankenstuecke, line 70:
```suggestion
    first_hand_authorship = forms.ChoiceField(choices = authorship_choices, widget = forms.RadioSelect(), required=True)
```

gedankenstuecke (approved):
Looks great now, and happy to have this merged once the `first_hand_authorship` is made `required=True` in the `forms.py` again! 🙂

gedankenstuecke (approved):

gedankenstuecke:
feel free to press the big green button @helendduncan 🚀

PR #582, opened 2023-09-04
Seed db for user testing

gedankenstuecke:
For the user testing we'd ideally have some existing, public experiences so that folks don't look at an empty page when browsing experiences. 

This PR adds a simple management command to import example experiences into the database from a user specified CSV file – alongside a CSV file with example experiences that the community wrote.


server/apps/main/management/commands/seed_db.py
```diff
@@ -0,0 +1,47 @@
...
+        # 4. (optional entry): which trigger label applies
+
+        with open(file_name, newline='') as csvfile:
+            reader = csv.reader(csvfile, delimiter=',', quotechar='"')
+            for i,row in enumerate(reader):
+                if i > 0:
+                    pe_data = {
+                        "title_text": row[0],
+                        "experience_text": row[1],
+                        "difference_text": row[2],
```
helendduncan, line 41: Might have to add in authorship here if this merges after the other PR

helendduncan (approved):
Command to seed database = `docker exec -it autspaces-web-1 python manage.py seed_db -f server/apps/main/tests/fixtures/example_experiences_seed.csv`

PR #585, opened 2023-09-08
User Testing: Restore DB to initial conditions

### LINKED ISSUE #584: Restore test db
### Summary
Given user testing is ongoing - Bastian has set up a placeholder database with sample stories. User testers will then use a test OpenHumans account to sign into the platform and add stories.

After the user test the following should happen

- [x] Remove the user added stories from the db
- [x] Remove the user added stories from the test OH account
- [x] Remove the test OH account
- [x] Remove the seeded stories

An extension after PR #583 

### Deadline 

September 2023

### What needs to be done?

See summary
### Who can help?

Anyone
---

helendduncan:
closes #584 
Management command to delete the user and their OH data and then remove the seeded stories from the DB
Edited `seed_db.py` to remove its analogous functions.

### DIFF
--- a/server/apps/main/management/commands/seed_db.py
+++ b/server/apps/main/management/commands/seed_db.py
@@ -17,17 +17,6 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         file_name = options["file"]
         
-        # delete existing OH members & PublicExperiences, 
-        # should reset DB except admin users before doing import
-
-        oh_members = OpenHumansMember.objects.all()
-        for ohm in oh_members:
-            ohm.delete()
-
-        pes = PublicExperience.objects.all()
-        for pe in pes:
-            pe.delete()
-
         # create OH member for public experience import
         data = {"access_token": 'foo',
                 "refresh_token": 'bar',
--- a/server/apps/main/management/commands/unseed_db.py
+++ b/server/apps/main/management/commands/unseed_db.py
@@ -1,6 +1,7 @@
 from django.core.management.base import BaseCommand
 from server.apps.main.models import PublicExperience
 from server.apps.users.models import UserProfile
+from django.contrib.auth.models import User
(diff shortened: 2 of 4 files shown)


gedankenstuecke (changes requested):
I think there's actually a way that's a lot easier to achieve the same thing. Which the original file in `seed_db` didn't take into account either, but I just noticed this:

`PublicExperience` objects use `on_delete=models.CASCADE`, so by iterating over all users and deleting them, the corresponding public experiences are deleted as well! 

Does that make sense?

helendduncan:
I like it - Frustratingly the test fails - I am assuming that there are multiple calls to an external function and the cassette only allows one - but I can't quite see how that would be the case as there should only be one deletion - the user's profile

gedankenstuecke:
yeah, it looks like only `oh_id=12345678` is the one that should be deleted, not the other one 🤔

gedankenstuecke:
@helendduncan I [found the problem in the docs of `django-open-humans`](https://django-open-humans.readthedocs.io/en/latest/modules/configuration.html#environment-variables)!

The default for `OPENHUMANS_DEAUTH_ON_DELETE` is `True`, which means that if we end up deleting a `User` object in our DB, we try to also ID the `OpenHumansMember` object associated and then send through a message through to Open Humans to please de-auth that user on their end as well. 

With our made-up user `999999999` that of course doesn't work, as the access_token is made-up. To fix this `OPENHUMANS_DEAUTH_ON_DELETE = False` needs to be added to the config that is used to run the test!

gedankenstuecke:
I'd actually suggest just adding that flag into the `development.py` setting as I think it's quite reasonable to have this disabled in both development & testing (and testing also uses those settings atm)

gedankenstuecke (approved):
That looks great, thanks @helendduncan! Also great that the tests now pass 🚀

PR #593, opened 2023-09-13
send explicit deauth if not default

### LINKED ISSUE #589: Explicitly send de-auth to OH when deleting local user
### Summary

So far we automatically send a "deauth"-request to Open Humans when deleting a local user object. This was not really a conscious decision on our behalf, but is rather the default that `django-open-humans` implements. 

In #585 we noticed this and for local testing and development might want to override this. At the same time we should make sure for the actual "user deletion" option that this turned on. 

### What needs to be done?

- [ ] Add a line in the `delete_user` function that runs the de-auth for the OH before account deletion

See https://github.com/alan-turing-institute/AutSPACEs/blob/fb63201cf4335f770598a9fb05b7bfacfad776c9/server/apps/users/helpers.py#L58 

### Who can help?

* 
* 

---

gedankenstuecke:
Closes #589

### DIFF
--- a/server/apps/users/helpers.py
+++ b/server/apps/users/helpers.py
@@ -75,7 +75,8 @@ def delete_user(user, delete_oh_data):
         ohmember.delete_all_files()
 
     # de-auth OH member if not doing so by default
-    if settings.OPENHUMANS_DEAUTH_ON_DELETE == False:
+    deauth_on_delete = getattr(settings, 'OPENHUMANS_DEAUTH_ON_DELETE', True)
+    if deauth_on_delete == False:
         deauth_oh_member(ohmember)
 
     # Delete the actual user
--- a/server/apps/users/helpers.py
+++ b/server/apps/users/helpers.py
@@ -1,6 +1,9 @@
 from django.contrib.auth.models import User
 from .models import UserProfile
 from server.apps.main.models import PublicExperience
+from django.conf import settings
+import requests
+from urllib.parse import urljoin
 
 def user_profile_exists(user):
     """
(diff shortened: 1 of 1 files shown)


gedankenstuecke:
@llewelld I've invited you for the review as you raised the point in the first place 🙂

server/apps/users/helpers.py
```diff
@@ -71,6 +74,17 @@ def delete_user(user, delete_oh_data):
...
+    # de-auth OH member if not doing so by default
+    deauth_on_delete = getattr(settings, 'OPENHUMANS_DEAUTH_ON_DELETE', True)
+    if deauth_on_delete == False:
+        deauth_oh_member(ohmember)
+
     # Delete the actual user
     user.delete()
 
+def deauth_oh_member(oh_member):
+    deauth_url = 'https://www.openhumans.org/api/direct-sharing/project/remove-members/'
```
llewelld, line 86: Following the example in `server/apps/main/tests/jh_tests.py` it would seem to make sense to use `OPENHUMANS_OH_BASE_URL` here too.
```
from openhumans.settings import openhumans_settings
OH_BASE_URL = openhumans_settings['OPENHUMANS_OH_BASE_URL']
...
    deauth_url = urljoin(OH_BASE_URL,
                         '/api/direct-sharing/project/remove-members/')
...
```

llewelld (changes requested):
Works well and the code looks good; I've made one suggestion.

Just an observation, but the manual deauth feels like functionality that should be exposed by `django-open-humans`, separate to the automatic deauth.

server/apps/users/helpers.py
gedankenstuecke: Great point, I had totally forgotten that we have that setting!

gedankenstuecke:
Yeah, I have to admit I was surprised myself that we don't have that call as part of the `django-open-humans` package! Do you want to make an issue in the GH repo of the project or should I? :)

llewelld:
> Yeah, I have to admit I was surprised myself that we don't have that call as part of the `django-open-humans` package! Do you want to make an issue in the GH repo of the project or should I? :)

As it happens I started creating a PR; I'll submit it shortly.

gedankenstuecke:
Awww, thanks!

llewelld (approved):
Still works well for me, so also LGTM.

gedankenstuecke:
Great, thanks! 🚀

PR #595, opened 2023-09-18
User Testing Response - My Stories page

### LINKED ISSUE #591: User testing: My Stories page improvements
### Summary

1. Message box should show latest "moderation changes"with a human-readable time (e.g. "Your story X was approved 7 days ago").
2. "Add Story" button.
3. In Process/Moderated numbers should be real.
4. Add 'All stories' tab.
5. Remove 'Triggering Labels' column.

### What needs to be done?

- [x] Message box.
- [x] Add Story button.
- [x] Working numbers.
- [x] Add 'All stories' tab.
- [x] Remove 'Triggering Labels' column.

### Who can help?

* Anyone willing to hack the page.

---

helendduncan:
In response to feedback from user testing. Closes #591

My Stories page has been updated.
The in-process, moderated, and messages boxes have been updated.
An "All Stories" tab has been added, and an "Add Story" button at the bottom of the page

Additionally the first two points of #616 are addressed in this PR


gedankenstuecke (approved):
LGTM, the functionality works as expected!

server/apps/main/views.py
```diff
@@ -458,6 +467,14 @@ def my_stories(request):
 
         # For each category, filter stories, create pagination and add continuous numbering
 
+        # All stories
+        paginator_all = Paginator(
+            files, items_per_page,
+        )
+        all_stories = paginate_stories(request, paginator_all, "page_all")
+        # Numbering for all stories different to all other tabs 
+        all_stories_numbering = list(range(all_stories.start_index(), all_stories.start_index()+min(items_per_page, len(all_stories.object_list))))
```
llewelld, line 476: This already works well as it is, but I'm wondering what the reason is that you can't pass `all_stories` into `number_stories()` like the others here. If you could it looks like it might simplify the `story_table.html` template.

server/apps/main/views.py
```diff
@@ -450,6 +452,13 @@ def my_stories(request):
         context = {"files": files}
         context = reformat_date_string(context)
 
+        print(request.user.openhumansmember)
```
llewelld, line 455: This looks like a debug line; should it be removed?

server/apps/main/helpers.py
```diff
@@ -657,4 +657,46 @@ def experience_titles_for_session(files):
...
+def number_by_review_status(files):
+    """
+    Return a dictionary of review status: number for the My Stories page
+    """
+    list_status = ["approved", "not_reviewed", "rejected", "in_review"]
+    status = {s: 0 for s in list_status}
+
+    for f in files:
+        s = f['metadata']['data']['moderation_status']
+        status[s.replace(" ","_")] = status[s.replace(" ","_")] + 1
```
llewelld, line 671: I like the way you've managed to efficiently roll all four types into a single loop.

llewelld (approved):
This works really nicely and I especially like the way the dashboard now populates. I added a couple of comments, but this already works well so I'm also happy for it to go in as it is.

llewelld:
It looks like the branch has diverged from main, so for the sake of maintaining a clean history, I'd suggest rebasing before merging.

server/apps/main/views.py
helendduncan: From my playing with things - you can only number a story once. So each story already has a number, and if you have one of each type you get 4 number 1s in the list. Or the "All Stories" numbering would overwrite the individual tab numbering. But this isn't something I really researched

llewelld: Oh, that's interesting. I guess the story dictionaries contain references to rather than copies of the stories; not necessarily a bad thing.

llewelld: In that case I can see it would take some refactoring to work around this; better to leave it as it is.

Issue #592, opened 2023-09-12
User testing: Navigation bar


llewelld:
### Summary

Changes to the menu bar along the top.

AutSPACEs
1. View Stories
2. Share Storeis (when logged in)
3. My Stories (when logged in)
4. Moderation -> Pending, Approved, Rejected (tabs?) (when logged in)
5. Learn More -> What is Autism, About Us, CoC, Help

### What needs to be done?

- [ ] Update the menu bar to reflect the structure above.

### Who can help?

* Anyone who wants to hack the menu bar.

---

llewelld:
Would it make sense to change the title on the 'Shared Stories' page to 'View Stories' to match the menu item?

![Screenshot 2023-09-25 at 17 49 48](https://github.com/alan-turing-institute/AutSPACEs/assets/1446122/e2d02f65-e985-4fd2-9414-dc17765ba5f3)

gedankenstuecke:
Yes, I think that would be great! :)

gedankenstuecke:
<img width="527" alt="Screenshot 2023-10-05 at 21 47 31" src="https://github.com/alan-turing-institute/AutSPACEs/assets/674899/4d5a9648-262d-4ef9-b2d1-8493227c5abd">

Looking at the mobile/collapsed view, the "account" sub-menu looks like it's indented a bit for some reason. I'm not sure if this was always the case or is a result of recent changes, but should we try to get it into the same "depth"? 🙂

PR #597, opened 2023-09-25
Rearrange the navbar

llewelld:
Changes the structure of the navigation menu bar at the top of every page, following feedback from user testing.

Contributes to #592.

### DIFF
--- a/server/apps/main/templates/main/experiences_page.html
+++ b/server/apps/main/templates/main/experiences_page.html
@@ -13,7 +13,7 @@
   <div class="container">
     <div class="row">
       <div class="col-lg-12 text-left">
-        <h1 class="big-heading">Shared Stories</h1>
+        <h1 class="big-heading">View Stories</h1>
       </div>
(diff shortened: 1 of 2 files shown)


llewelld:
An image capture of the new structure:

image

llewelld:
Issue #592 also has the Moderation drop down turned into tabs on the Moderation page, but I'd quite like to separate this off into its own PR. I've therefore moved this out of draft.

gedankenstuecke (approved):
This looks good to me. 

I think we had also considered to keep a "share stories" point around even when not logged in, which would take people to a landing page explaining how to contribute, but that might be a bigger change that's best left for another PR? :)

llewelld:
Thanks for the approval @gedankenstuecke. There are now a couple of changes to make on top of this (changing "Share Stories" to "Add Story" and retaining it when logged out). I'll merge this so we can build on these changes in a separate PR.

PR #623, opened 2023-09-27
Switch moderation pages to use tabs

llewelld:
Switches the three moderation pages (pending, approved, rejected) to use tabs on a single page.

Contributes to #592.

### DIFF
--- a/static/js/store-tab.js
+++ b/static/js/store-tab.js
@@ -1,18 +1,21 @@
 // This file is used to store the last active tab in the browser's local storage
 $(function() {
-    // when a tab is clicked
-    $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
-      // save the latest tab using a cookie
-      localStorage.setItem('lastTab', $(this).attr('id'));
-    });
+  // generate a simple unique deterministic ID for the page
+  const pageid = window.location.pathname.replaceAll('/','_')
 
-    // get the last tab if it exists
-    var lastTabStories = localStorage.getItem('lastTab');
-    
-    if (lastTabStories) {
-      $('#' + lastTabStories).tab('show');
-    } else {
-      // Set the first tab to active if no active tab exists
-      $('a[data-toggle="tab"]:first').tab('show');
-    }
+  // when a tab is clicked
+  $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
+    // save the latest tab using a cookie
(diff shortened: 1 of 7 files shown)


llewelld:
Mostly reusing @mastoffel's nice tab code.

static/js/store-tab.js
```diff
@@ -1,18 +1,21 @@
...
-      // Set the first tab to active if no active tab exists
-      $('a[data-toggle="tab"]:first').tab('show');
-    }
-  });
\ No newline at end of file
+  // when a tab is clicked
+  $('a[data-toggle="tab"]').on('shown.bs.tab', function (e) {
+    // save the latest tab using a cookie
+    localStorage.setItem('lastTab' + pageid, $(this).attr('id'));
+  });
```
gedankenstuecke, line 10: This is really nice, I wondered how it worked and that's a really nice touch. I wonder if that might also be useful for the trigger toggles/view experiences page? also flagging @helendduncan who I think had looked at that at some point to avoid overloadingt the URL with parameters?

gedankenstuecke (approved):
Well done @llewelld, that works really well and was a pleasure to test <3

static/js/store-tab.js
llewelld: Yes, agreed! (Props to @mastoffel for coming up with the approach!).

helendduncan:
I think this works very nicely and is intuitive. My only comment would be that the moderation table spans the page which is different to say the my stories tab. If it were an issue would a placing the table in a container fix it?

helendduncan (approved):

gedankenstuecke:
> I think this works very nicely and is intuitive. My only comment would be that the moderation table spans the page which is different to say the my stories tab. If it were an issue would a placing the table in a container fix it? Happy to approve this PR either way!

That's a great catch! I think that's maybe a larger thing we should check for consistency, I remember adding the main container divs on some of the templates at some point but it's probably still missing on some!

llewelld:
> I think this works very nicely and is intuitive. My only comment would be that the moderation table spans the page which is different to say the my stories tab. If it were an issue would a placing the table in a container fix it? Happy to approve this PR either way!

Thanks! Good that you noticed. But I'm not seeing this myself/am misunderstanding the problem. Perhaps you can show me the issue when we meet next week? That would be really helpful.

gedankenstuecke:
Sounds like a great plan to me, let's do that on Tuesday 👍

llewelld:
Rebased.

PR #624, opened 2023-09-28
Update django-open-humans + remove custom deauth

gedankenstuecke:
Thanks to @llewelld's upgrade of the `django-open-humans` package the `OpenHumansMember` objects now have a working `.deauth()` function associated, making it unnecessary to duplicate that functionality on our end. 

This PR updates our dependencies to use the updated `django-open-humans` package and removes our duplicate implementation.


llewelld (approved):
I gave it a quick test and it seemed to work great. Thanks for the changes @gedankenstuecke.

llewelld:
I had to restart the docker conainer to get this change to work through fully (not a surprise or problem in any way, just something to note).

gedankenstuecke:
Ah yes, that makes sense, as I guess the bump in versions requires building a new container and is the kind of thing that the "on the fly-reloading" of Django doesn't cover "install new libraries inside the container" 😂

Issue #586, opened 2023-09-12
User Testing: Update front page


helendduncan:
### Summary

During user testing the following comments/issues were noted with the from page

- [x] No links to reading experiences
- [x] No links to writing experiences (or logging in)
- [x] Get involved and what autism is can be removed from front page
- [x] Add description of the platform
- [x] Increase/share/advise -> "our goals" or similar
- [x] How to use platform -> "Get Involved - use yellow" Reading stories + Contributing stories (with links) (- "Read More")
- [x] Shared stories: Replace placeholder stories with published stories (? pull from DB - set env variable to pick from selected uuids?) If not use current placeholder/seed stories
- [x] Original get involved (GitHub links etc) can go
- [x] Original what autism is can go
### Deadline 
Mid October 2023

### What needs to be done?
See checklist above

### Who can help?

---

llewelld:
In relation to the **"Increase/share/advise -> "our goals" or similar"** item, there was a comment made during user testing that relates to this and which it might make sense to tackle at the same time:

"Unsure about meaning of "advise" text - thinks it should link to "advice" page and potentially link to other websites"

This relates to the wording of the text on the front page that says "Advise - organisations how to create better environments for autistic people". We should consider tweaking the wording (e.g. to remove the term "Advise") to try to avoid this confusion.

Note @GeorgiaHCA's PR #581 updates the copy on the page, so it should be checked whether this already fixes the issue.

PR #626, opened 2023-10-04
Restructure home page

llewelld:
Updates te homepage layout and content following user testing and so that all of the elements work correctk (e.g. links to other parts of the site).

Contributes to #586.

### DIFF
--- a/server/apps/main/templates/main/home.html
+++ b/server/apps/main/templates/main/home.html
@@ -4,13 +4,11 @@
 
 {% block body_attributes%} data-spy="scroll" data-target="#content-bar" {% endblock %}
 
-  {% block content %}
+{% block content %}
 
-<!--Navigation-->
+<!-- Introduction of AutSPACEs -->
 <section id="title">
-<!--  Navigation-->
-   <div class="container-fluid">
-     <!--Introduction of AutSPACEs-->
+  <div class="container-fluid">
     <div class="row intro">
       <div class="col-lg-6 text-left">
         <h1 class="big-heading">AutSPACEs</h1>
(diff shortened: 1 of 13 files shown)


llewelld:
This should be ready so I'm removing the draft tag.

The implementation is suboptimal: the uuids for real stories should go in the database rather than the carousels.json file. But it felt like there were probably more pressing things to work on, so I've created task #627 to fix this.

llewelld:
Closes #586.

server/apps/main/views.py
```diff
@@ -116,11 +117,15 @@ def index(request):
     """
     if request.user.is_authenticated:
         oh_member = request.user.openhumansmember
+
+        stories = get_carousel_stories()
```
gedankenstuecke, line 121: For the laughs I'll put this here what I had written as the original top-level comment for the PR: 

> It might be because my brain is a bit fried, but I think I struggle to understand how the carousel bit is supposed to work in practice, in particular how it prioritizes what to show. In my local dev setup (using the `carousel.json` from this PR without any changes but having some approved local stories) the carousel is just completely empty. Is that the expected behaviour? 🙂 

Having had a second look, this bit is just put at the wrong place. Currently we only `get_carousel_stories()` for users that are authenticated, and that's why my logged-out session didn't load any stories 😂

gedankenstuecke: I should add: Logged-in the carousel works as I would kind of have expected, it shows the pre-existing data from the carousel.json

server/apps/main/templates/main/home.html
```diff
@@ -20,135 +18,156 @@ <h1 class="big-heading">AutSPACEs</h1>
...
+      if autistic people had more of a say in how public and private spaces are
+      used and designed.</p>
 
-  <div class="col-lg-4 value-box">
-  <h3 class="value-heading">Advise</h3>
-  <p class="value-text">organisations and policy-makers for real-world impact</p>
-    </div>
+      <p>All autistic people are different. We want to collect together lots of
+      stories from autistic people, so that we can better understand how
+      sensory processsing differences affect the way autistic people navigate
```
gedankenstuecke, line 47:
```suggestion
      sensory processing differences affect the way autistic people navigate
```

server/apps/main/templates/main/home.html
```diff
@@ -20,135 +18,156 @@ <h1 class="big-heading">AutSPACEs</h1>
...
 </section>
-<!--How to Use the Platform-->
+
+<!-- How to Use the Platform -->
 <section id="platform_usage">
   <div class="row">
-      <div class="col-lg-12 platform-title">
-      <h2 class="section-heading">How to Use The Platform</h2>
+    <div class="col-lg-12 platform-title">
+      <h2 class="section-heading">How to use the platform</h2>
```
gedankenstuecke, line 66: I don't remember on top-of-my-head what the user testing results where on this, but I do feel like we can potentially reorder and simplify this bit? I'd say the order 

1. View
2. (Learn)
3. Register
4. Share

could be an easier one, I don't think the edit necessarily needs to appear, and even the "learn" could go into "view"? 

Tagging @GeorgiaHCA in case she has a suggestion!

server/apps/main/templates/main/home.html
```diff
@@ -20,135 +18,156 @@ <h1 class="big-heading">AutSPACEs</h1>
...
-        <div class="col-md-5">
-        <h3 class="experience-heading">Eating in a Restaurant</h3>
-        <p class="experience-content">...Waiting for the food felt like some kind of marathon and I just got so overwhelmed. But now that I’ve
-          realised hunger exacerbates my sensory differences I know how to cope. I try to make sure I’ve eaten something before I go out and a while back I bought some earplugs which are great for small and bustling places.. </p>
-        <em small class="experience-content">Sdy</em>
-         </div>
+          {% if story.image %}
+          <div class="experience-half w-50">
+          <div class="experience-image-box w-100">
+          <img class="experience-image d-block w-100" src="{% static "/images/" %}{{ story.image }}" alt="First slide" />
```
gedankenstuecke, line 150: If we think that the animations/images benefit users that require alt-text, then I think it would be good if we could also add the `alt` text into the `carousel.json` to have proper alt text instead of the place holder.

Having said that, I think we might just want to drop the alt-text, as I think the general recommendation is that if animations are just "visual fluff" and not necessary for the understanding then one should skip the alt-text as people using screenreader don't benefit from having these "flavour texts" read out.

gedankenstuecke (commented):
Really great work @llewelld! Barring one bug and one typo I think these are now mostly comments that are open for discussion whether to do any changes on these or not!

server/apps/main/views.py
llewelld: Nice catch! This should be fixed now.

server/apps/main/templates/main/home.html
llewelld: Oh, also nicely caught. Thanks. This should also now be fixed.

server/apps/main/templates/main/home.html
llewelld: Ah, I'd planned to do something with these bug forgot. Thank you.

I've removed the alt tag text for now, but will happily put something in the json file for them if that's prefered?

gedankenstuecke: For now I'd say let's just go with the "no alt text" to not _spam_ screenreaders with it!

gedankenstuecke (approved):
LGTM! I'd be happy to have that merged in and then we can discuss additional changes to the front page outside this PR to not have the perfect be the enemy of the good!

Issue #576, opened 2023-08-29
[BUG]: Help page contents list overlaps main text and doesn't scroll


llewelld:
### Describe the bug

On the help page there is a Contents list shown on the page. In desktop mode this positions itself on the left of the page with the main help text to the right.

There seem to be two potential issues with the contents list:
1. [x] If there are too many items in the list so that it extends beyond the bottom of the page, it's not possible to access the items that are cut off (e.g. no way to scroll the list).
2. [x] If the width of the page is too small the main help text will overlap the contents list.

### Steps to reproduce

1. Select 'Help' in the menu along the top of the site.
2. Either reduce the height/width of the browser window until the errors appear; or
3. Select the Browser's [Responsive Design Mode](https://superuser.com/a/1292454) and adjust the viewport height and width to give the same result.

### In which environment are you using AutSPACEs?

MacOS and Firefox.

gedankenstuecke:
The user testing found this to be a problem as well, in particular when the viewport is too narrow (e.g. on an iPad in portrait orientation). 

There's two aspects to this: 

1. If the viewport is too narrow (left-to-right) to display side-by-side, the navigation should turn into a traditional table of contents on the top
2. If the viewport is not tall enough, it could either become a scrollable menu on the left or also go to the top

gedankenstuecke:
![Example screenshot of the issue](https://user-images.githubusercontent.com/674899/270681521-05dd0d1b-4649-463d-905e-8fb45b648ab3.PNG)

see screenshot above for an example of the problem

llewelld:
The problem appears to be that the `content-bar` (which contains the menu items) has no height. As a result, although Bootstrap tries to position the right hand columns below it when the width of the page gets too small, it positions it on top instead.

llewelld:
The lack of height seems to be caused by the fact that the bar has its position fixed:
```css
#content-bar {
 padding: 2% 0 0% 5%;
 position: fixed;
}
```
Removing the `position: fixed;` "fixes" the issue, but also makes it so that the menu scrolls with the page (rather than remaining on the page).

It's possible this can be fixed by using the `@media width` declarations (i.e. make it fixed when the page is wide enough, relative otherwise).

gedankenstuecke:
[...]

PR #628, opened 2023-10-05
Avoid help page menu overlapping text

llewelld:
When the width of the viewport is reduced below 992 pixels the menu along the left hand edge overlaps the text on the page. Using Bootstraps columns it should jump above the text, but this is prevented because the positioning of the menu is set to fixed (so that it stays on screen when the page is scrolled).

This change fixes this by creating a media selector that makes the menu position relative when the width of the page reduces below 992 pixels. An invisible column is also added to avoid the need to place the menu inside another column, and to prevent the text from shifting to the left when the menu position is set to fixed.

The footer position is also set to relative so that it goes over (rather than under) the menu to avoids a nasty overlap when viewport height is small.

Contributes to #576.

### DIFF
--- a/server/apps/main/templates/main/about_us.html
+++ b/server/apps/main/templates/main/about_us.html
@@ -12,8 +12,11 @@
 <section id="content">
   <div class="container-fluid">
     <div class="row">
-      <div class="col-3">
-        <div class="list-group col-lg-3 cotent-list" id="content-bar">
+      <div class="col-lg-3">
+        <div class="col-lg-3" id="content-bar-filler">
+        <!-- Space filler to avoid the main text shifting leftwards -->
+        </div>
+        <div class="list-group col-lg-3 content-list" id="content-bar">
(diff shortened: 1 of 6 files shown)


llewelld:
Addresses part of #576, but doesn't address the issue of the manu not scrolling when the viewport height is too small.

gedankenstuecke (approved):
That works super well, amazing job again @llewelld!

llewelld:
I've added a commit to move the menu to the top in case the viewport height isn't large enough to accommodate it.

It's not ideal. For example it takes the height of the menu on first render, which means it won't be responsive to changes in font size or other things that affect the height of the menu. But it's probably a reasonable starting point.

gedankenstuecke (approved):
I just tried and I think it works really well (barring changes in font size), but otherwise it's really nicely responsive! I'd be happy to use this as a solution for now!

PR #630, opened 2023-10-10
fix accidental duplicate

gedankenstuecke:
fixes problem introduced in #629


llewelld (approved):
Properly tested this time and LGTM!

Issue #605, opened 2023-09-26
Make platform more comfortable for R/G colour-blind users 


GeorgiaHCA:
### Summary

Currently, the platform uses variants of reds and greens. These are visible to colour blind users but are still uncomfortable and can make eyes tired. We need to adjust the colour palette so it is more comfortable for people with R/G colour blindness. 

### What needs to be done?

- [ ] Adjust colour palette on all pages 
- [ ] Test again with people with R/G colour blindness

gedankenstuecke:
The colour red was generally commented on as potentially problematic as it is a warning colour

GeorgiaHCA:
Multiple users didn't like the red colour - it was considered too much of a "warning colour"
Other users felt that there were too many colours and this was overwhelming
-- Suggest restricting the colour palette so their are fewer colours, focussing on the turquoise as a main colour

gedankenstuecke:
Partially addressed with the dark mode in #604, colours will be addressed in #667

Issue #603, opened 2023-09-26
Have a dark theme available


GeorgiaHCA:
### Summary

AutSPACEs is not accessible to people with some visual sensory sensibilities. Having a dark theme would allow more users to feel comfortable with the website. 

### What needs to be done?

- [ ] Give users the option of changing the settings to include a dark theme - allow this to be easily adjusted and put it in a clear place so that it is easy to find 

### Who can help?

Please put suggestions of where and how to include the dark theme option in the comments below

* 
* 

---

GeorgiaHCA:
Oops multiple similar issues

PR #631, opened 2023-10-10
Migrate to Bootstrap 5x

gedankenstuecke:
This will potentially be a bit of work as a lot of declarations have changed, but it would be good to migrate now rather than later in order to avoid being too outdated. This will also help with a lot of the accessibility improvements identified during the user tests (e.g. #605 & #603). 

- [x] Change main CSS/JS files
- [ ] Check if there are more CSS/JS files that need to be changed
- [x] Fix navbar 
- [x] Fix broken JS on moderation index page (tabs not working)
- [x] Fix broken JS on moderation page (selecting moderation reason not working)
- [x] Fix broken JS on "my stories" page (tabs not working)
- [x] Fix "badges" on My Stories page
- [x] Fix "remember tabs" for moderation page & my stories page
- [x] Fixen toggles (all converted back to regular check boxes) 
- [x] Fix broken "view stories" page (expanding stories does not work) 
- [x] Fix "scrollspy" pages (about, CoC, autism, …) 
- [x] Fix any pages with a form help texts don't properly display
- [x] Colours of headings on about/autism/coc/help pages are off?
- [x] Fix carousel on `home.html`

### DIFF
--- a/server/apps/main/templates/main/about_us.html
+++ b/server/apps/main/templates/main/about_us.html
@@ -18,7 +18,7 @@
         <!-- Space filler to avoid the main text shifting leftwards -->
         </div>
         <div class="list-group col-lg-3 content-list" id="content-bar">
-          <h5 class="list-group-item content-title">Contents</h5>
+          <div class="list-group-item content-title fw-bold">Contents</div>
           <a class="list-group-item list-group-item-action  d-flex justify-content-between align-items-center"
              href="#about-autspaces">About AutSPACEs<span class="badge badge-pill"></span></a>
           <a class="list-group-item list-group-item-action  d-flex justify-content-between align-items-center"
--- a/server/apps/main/templates/main/about_us.html
+++ b/server/apps/main/templates/main/about_us.html
@@ -19,7 +19,7 @@
         </div>
         <div class="list-group col-lg-3 content-list" id="content-bar">
           <h5 class="list-group-item content-title">Contents</h5>
-          <a class="list-group-item list-group-item-action  d-flex justify-content-between align-items-center active"
+          <a class="list-group-item list-group-item-action  d-flex justify-content-between align-items-center"
              href="#about-autspaces">About AutSPACEs<span class="badge badge-pill"></span></a>
           <a class="list-group-item list-group-item-action  d-flex justify-content-between align-items-center"
              href="#platform-usage">Platform Usage<span class="badge badge-pill"></span></a>
--- a/server/apps/main/templates/main/about_us.html
+++ b/server/apps/main/templates/main/about_us.html
@@ -39,7 +39,7 @@
       </div>
       <div id="text-container" class="about-content col-lg-9" data-bs-offset="0" style="padding: 2% 8%">
         <div class="about-content-section">
-          <h5 class="text-info" id="about-autspaces"><em></em>About AutSPACEs</h5>
+          <h5 class="heading-blue" id="about-autspaces"><em></em>About AutSPACEs</h5>
           <p>The “Aut” in AutSPACEs refers to Autistic people, and the “SPACEs” is an acronym which refers to exploring
             Sensory Processing for Accessible Community Environments. This captures the aim of AutSPACEs which is to make
             spaces better for autistic people by learning about sensory processing and its impact.</p>
@@ -67,7 +67,7 @@
           </ul>
         </div>
         <div class="about-content-section">
-          <h5 class="text-info" id="platform-usage"><em></em>Platform Usage</h5>
+          <h5 class="heading-blue" id="platform-usage"><em></em>Platform Usage</h5>
           <h6 class="platform-list-title">1. Register with Open Humans</h6>
           <p>When you register you create a private and anonymous account with Open Humans. We work with Open Humans to
             make sure you have control over your data and that it is safe.</p>
@@ -92,7 +92,7 @@
             submitted.</p>
         </div>
         <div class="about-content-section">
-          <h5 class="text-info" id="explore-autspaces"><em></em>Explore AutSPACEs</h5>
+          <h5 class="heading-blue" id="explore-autspaces"><em></em>Explore AutSPACEs</h5>
           <p>You do not need an account to explore AutSPACEs and read other people’s stories.</p>
 
           <p>If you are autistic, you may find other people who have similar sense experiences to you. You can discover tips
@@ -102,7 +102,7 @@
             them how their senses affect them.</p>
         </div>
         <div class="about-content-section">
-          <h5 class="text-info" id="data-usage"><em></em>Data Usage</h5>
+          <h5 class="heading-blue" id="data-usage"><em></em>Data Usage</h5>
           <p>We will <b>never</b> sell your data or share it with anyone without your consent.</p>
 
           <p>We believe that your data should belong to you, so it’s your choice what to do with it.</p>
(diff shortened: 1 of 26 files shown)


gedankenstuecke:
Memo for myself: The scrollspy things are half-working now, they work once and then not when scrolling up again 🙈

gedankenstuecke:
I looked into the changes in colours for some texts, and those are a change in how BS treats it's colours (e.g. the info is now a lot more cyan than the muted colour it used to be). I can go ahead and edit those in our own `main.css`

gedankenstuecke:
okay, those colours are fixed :)

gedankenstuecke:
Okay, I spent another 30 minutes on the scrollspy issue but to no avail 😂

llewelld:
This is amazing work @gedankenstuecke.

Concerning scrollspy, I wonder if you could be hitting [Bootstrap issue #36431](https://github.com/twbs/bootstrap/issues/36431). It looks like a known issue with 5.2.

I tested out your changes with Bootstrap 5.1.3 and found the scrollspy worked considerably better. Would 5.1.3 be enough, at least until the fix is released?

gedankenstuecke:
Yeah, I found that issue too and wondered if that's somehow related! 

The main trade-off between 5.1 and 5.2 is that only 5.2 comes with the dark mode options that at least partially motivated the wish to update bootstrap 🙈 If we wanted to stick with 5.2 we could either decide to up drop the scrollspy or replace it with a 3rd party option instead?

llewelld:
Yes I see; that's a really good point.

Even though it's frustrating, I'd say dropping it or replacing it would both be valid options. It shouldn't be hard functionality to replicate, but it might also be nice to be able to revert back to the BootStrap version once it's fixed (to avoid pulling in unecessary dependencies).

By the way, perhaps I'm misreading, but it looks like dark mode is 5.3 onwards (but still with the broken srollspy implementation).

gedankenstuecke:
Ah yes, great catch, it's 5.3+ for the dark theme (and it's the one this PR is currently based on, so that works in principle!) 

I'd personally not be too sad about seeing the scrollspy at least temporarily go, especially as it's already disabled (or at least not applicable) for the mobile view where the menu bar hops to the top of the page. 

Once I have some time I can look into what an alternative to bs's scrollspy could be!

gedankenstuecke:
I've removed the scrollspy for now, @llewelld it would be amazing if you had a look!

gedankenstuecke:
@llewelld I know your official allocation to this has ended, but you had mentioned that you had a look at this already. I just wanted to check if you still wanted to submit that review or if someone else should have a go at it? 🙂

llewelld:
Thanks for prodding me about this @gedankenstuecke. I do have a partial review that I've not had a chance to complete yet. Sorry for this. I still plan to submit it, but do of course feel free to have others take a look as well (if I miss my chance to add my comments, that's on me).

gedankenstuecke:
Thanks @llewelld, I'll see with @helendduncan if there's any reason that this would be urgent to get merged in!

helendduncan:
No hurry from my end @gedankenstuecke @llewelld I am only just back allocated on the project and my next task won't be held up by this, but am also happy to review if needed

gedankenstuecke:
Great, let's give @llewelld a bit more time to finish his review in that case! David, it's much appreciated that you're taking the time for this! 🙏

server/apps/main/templates/main/about_us.html
```diff
@@ -2,12 +2,13 @@
...
 {% block title %}AutSPACEs - About Us {% endblock %}
 
-{% block body_attributes%} data-spy="scroll" data-target="#content-bar" {% endblock %}
+{% block body_attributes%} {% endblock %}
 
 {% block content %}
 
 <script src="{% static 'js/menu-alignment.js' %}"></script>
 
+
```
llewelld, line 11: There's a small (0.5 rem) gap that's appearing between the heading menu item and the subsequent items; apparently added by the `<h5>`.

It looks okay actually, but I think is probably unintentional. Something like this could be used to remove it (added to `main.css`).
```
.list-group-item {
  margin-bottom: 0;
}
```

server/apps/main/templates/main/moderation_list.html
```diff
@@ -31,27 +33,29 @@
...
+                <div class="col-auto">
+                  <button class="btn btn-primary mb-2" type="submit">Search</button>
+                </div>
               </div>
-            </div>
-          </form>
+            </form>
+          </div>
         </div>
       </div>
```
llewelld, line 60: Something that becomes apparent in dark mode is that these cards have a border. It's barely visible in light mode but in dark mode it's really high contrast. But it highlights that there's no gap between the "Stories" card above and the "Review tabs" below.

It might be worth adding in some kind of gap, e.g. by adding in a top margin to the `moderation-table-container`:
```
.moderation-table-container {
  margin-top: 3%;
  margin-bottom: 3%;
} 
```

llewelld (commented):
This is really nice. Apart from the menu highlighting which we discussed elsewhere (and which I agree isn't a big loss) things look very similar to the Bootstrap 4 version. Really nice.

There were a couple of points that I can't add as code comments so I'll put them here instead.

I experienced quite a serious scroll glitch on the shared stories page, caused by the 'smooth scroll' behaviour interacting with the code that resets the scroll position after a search. I've no idea why the Bootstrap change might have caused this. At any rate, forcing instant scroll fixed it for me:

```diff
diff --git a/static/js/shared-stories-filtering.js b/static/js/shared-stories-filtering.js
index 0c68d0d..9850060 100644
--- a/static/js/shared-stories-filtering.js
+++ b/static/js/shared-stories-filtering.js
@@ -43,7 +43,7 @@ $("#single_trigger_warnings :checkbox").change(function(){
 
 document.addEventListener("DOMContentLoaded", function(event) { 
     var scrollpos = localStorage.getItem('scrollpos');
-    if (scrollpos) window.scrollTo(0, scrollpos);
+    if (scrollpos) window.scrollTo({left: 0, top: scrollpos, behavior: "instant"});
 });
 
 window.onbeforeunload = function(e) {
```

Completely separately, I found adding this helpful for testing. Although I appreciate this PR isn't supposed to fix everything about dark mode, this code will activate dark mode based on the OS/browser settings:
```diff
diff --git a/static/js/main.js b/static/js/main.js
index e69de29..f2d49b0 100644
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -0,0 +1,10 @@
+$(function () {
+  function selectTheme(media) {
+    $("html").attr("data-bs-theme", media.matches ? "dark" : "light");
+  }
+
+  const media = window.matchMedia('(prefers-color-scheme: dark)');
+  selectTheme(media);
+  media.addListener(selectTheme);
+});
+
```

Overall, the switch to Bootstrap 5 looks great to me.

server/apps/main/templates/main/about_us.html
gedankenstuecke: Great catch, I hadn't noticed. I've replaced the h5 with a more default declaration for the list view and made that bold instead, as I think that works well :)

gedankenstuecke:
Thanks a ton @llewelld, well spotted on those issues. I've implemented those fixes. And well done on getting the dark mode thing implemented right away!

@helendduncan feel free to also give it a look and see if it works for you too!

llewelld (approved):
Thanks for making the changes. This all looks great to me. Given the discussion about theming yesterday during the meetup, it sounds like your effort upgrading to v.5 will be very worthwhie.

gedankenstuecke:
Thanks for all the help @llewelld, and totally agree that it seems like it'll come in really useful to allow people to adapt the way we displays it! 

I'll merge it in now, so that we can see if there's any changes needed to #632 after it!

PR #632, opened 2023-10-11
Make it easier to add a story

### LINKED ISSUE #599: Make it easier to add a story
### Summary

Multiple testers found it difficult to know how to add a story on AutSPACEs and the instructions were quite hidden away. It would be useful to have an obvious add story button on the main homepage that would link to a diagram which takes you through the process and let you click on a registration link, and include other "add story" options throughout the platform. 

### What needs to be done?

- [x] Include an "add story" in nav bar for non-logged in users which takes you to the register or login page
- [x] Include on the register page information a clear user journey in the form of text and a diagram, plus a link to register, as well as clear text saying, "if you are already a user click here to log in" 
- [x] Include an obvious "add story" button on the homepage that would lead to the registration page if you are not logged in and the submit experience story if you are 
- [x] Include an "add story" button in the "view experiences" section of the website that would lead to the registration page if you are not logged in and the submit experience story if you are
- [x] Include a separate "add story" section on the nav bar for signed in users instead of it being part of a drop down menu

llewelld:
Adds "Add Story" buttons to various places.

Adds a process diagram to the registration page to try to clarify what the user needs to do to register an accont.

Fixes #599.

### DIFF
--- a/server/apps/main/templates/main/home.html
+++ b/server/apps/main/templates/main/home.html
@@ -53,11 +53,11 @@
   <div class="row">
     <div class="col">
       <div class="float-end">
-        <a href="{% url 'main:share_exp' %}" class="mr-3">
-          <button class="btn btn-outline-primary btn-home-about btn-lg mb-3" type="button">Add Story</button>
+        <a href="{% url 'main:share_exp' %}" class="mr-3 btn btn-outline-primary btn-home-about btn-lg mb-3">
+          Add Story
         </a>
-        <a href="{% url 'main:about_us' %}" class="mt-3">
-          <button class="btn btn-outline-primary btn-home-about btn-lg mb-3" type="button">Read More</button>
+        <a href="{% url 'main:about_us' %}" class="btn btn-outline-primary btn-home-about btn-lg mb-3">
+          Read More
         </a>
       </div>
     </div>
--- a/server/apps/main/templates/main/home.html
+++ b/server/apps/main/templates/main/home.html
@@ -51,10 +51,15 @@
     </div>
   </div>
   <div class="row">
-    <div class="col float-end">
-      <a href="{% url 'main:about_us' %}">
-        <button class="btn btn-outline-primary btn-home-about btn-lg float-end" type="button">Read More</button>
-      </a>
+    <div class="col">
+      <div class="float-end">
+        <a href="{% url 'main:share_exp' %}" class="mr-3">
+          <button class="btn btn-outline-primary btn-home-about btn-lg mb-3" type="button">Add Story</button>
+        </a>
+        <a href="{% url 'main:about_us' %}" class="mt-3">
+          <button class="btn btn-outline-primary btn-home-about btn-lg mb-3" type="button">Read More</button>
+        </a>
+      </div>
     </div>
   </div>
 </section>
@@ -78,11 +83,7 @@
     </div>
 
     <div class="col platform-box usage-box">
-      {% if user.is_authenticated %}
       <a href="{% url 'main:share_exp' %}">
-      {% else %}
-      <a href="{% url 'main:registration' %}">
-      {% endif %}
         <h3 class="usage-heading">
           Share
           <br/>
--- a/server/apps/main/templates/main/registration.html
+++ b/server/apps/main/templates/main/registration.html
@@ -24,8 +24,8 @@
       <p class="registration-text">OpenHumans is a non-profit organisation designed to help people access and understand their personal data,
         and to help them do and share things that use that data.
         You can find out more about Open Humans on <a href="https://www.openhumans.org/about/"
-                                                      style="color: #ffff" class="stretched-link">their website</a>.
-      <p class="registration-text">We will never sell your data, and we will never share it without your consent.
+                                                      style="color: #ffff">their website</a>.</p>
+      <p class="registration-text">We will never sell your data, and we will never share it without your consent.</p>
     </div>
   </div>
   <div class="row mt-3" style="color: #000;">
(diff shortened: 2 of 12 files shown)


server/apps/main/templates/main/home.html
```diff
@@ -51,10 +51,15 @@ <h2 class="section-heading">About AutSPACEs</h2>
...
-        <button class="btn btn-outline-primary btn-home-about btn-lg float-right" type="button">Read More</button>
-      </a>
+    <div class="col">
+      <div class="float-right">
+        <a href="{% url 'main:share_exp' %}" class="mr-3">
+          <button class="btn btn-outline-primary btn-home-about btn-lg mb-3" type="button">Add Story</button>
+        </a>
+        <a href="{% url 'main:about_us' %}" class="mt-3">
+          <button class="btn btn-outline-primary btn-home-about btn-lg mb-3" type="button">Read More</button>
+        </a>
```
gedankenstuecke, line 61: I think those button declarations are duplicating the links if defined like this! The one below should work: 

```suggestion
<a href="{% url 'main:share_exp' %}" class="btn btn-outline-primary btn-home-about btn-lg mb-3">
          Add Story
        </a>
        <a href="{% url 'main:about_us' %}" class="btn btn-outline-primary btn-home-about btn-lg mb-3">
          Read More
        </a>
```

gedankenstuecke: oh, and relatedly: currently when logged out the "add story" button always redirects to the front page, due to the redirect being set to that instead of the register page

server/apps/main/templates/main/registration.html
```diff
@@ -22,11 +24,78 @@ <h3 class="section-heading">Welcome to AutSPACEs Registration</h3>
...
+        </div>
+        <div class="card-body">
+          You're now ready to share your own story by selecting <b>Add Story</b> from the main menu
+        </div>
+      </div>
+    </div>
+  </div>
+  <div class="row mt-3" id="first">
+    <div class="col-lg-12">
+      <p><a href="https://www.openhumans.org/activity/autspaces/" target="_blank">
```
gedankenstuecke, line 85: Actually, this can also still be the login link, as the resulting page allows registering with OH and then takes you back to AutSPACEs at the end! It should probably jsut open in a new window/tab so that folks can keep this page around!

server/apps/main/views.py
```diff
@@ -257,7 +257,7 @@ def share_experience(request, uuid=False):
             )
 
     else:
-        return redirect("index")
+        return registration(request)
```
gedankenstuecke, line 260: weird, this new redirect does not work for me!

gedankenstuecke (changes requested):
Thansk @llewelld, overall that looks really great! And actually the diagram on the register page is really lovely and not bad at all! 🚀 

I left a few comments here and there, but it's all minor things I think! 

One thing I wondered: Should we try to get #631 merged in first, just to make sure that the added frontend declarations aren't immediately outdated?

server/apps/main/templates/main/registration.html
llewelld: Okay, I'll change it back. Does OpenHumans use the redirect header to figure out the project then? I don't know of a good way to test this while the project isn't officially active.

server/apps/main/views.py
llewelld: That is weird. What's the behaviour you're seeing? Does it go back to the home page instead?

The expected behaviour is that it doesn't redirect at all, but rather serves the Registration Page html content while retaining the `add_story` url.

server/apps/main/templates/main/home.html
gedankenstuecke: Yeah, I think every time I noticed that double declaration I tried fixing it, but I'm sure there's some places left! 😂

server/apps/main/views.py
gedankenstuecke: Yeah, so for me it just took me back to the `home` page instead of loading the registration page!

gedankenstuecke: Strangely, I can't reproduce this behaviour now, instead it correctly redirects.  🤔 Maybe there was just some caching problem or a problem that resolved itself by completely restarting the docker! so nevermind on this one!

server/apps/main/templates/main/registration.html
llewelld: Oh, now I get it. You're saying use the same link for both login and registration. It's conceptually a bit odd to have two buttons for this, but I get the motivation. That makes sense now.

server/apps/main/templates/main/home.html
llewelld: They should all be switchd now; at least the ones I could find.

server/apps/main/templates/main/registration.html
llewelld: This is now done.

llewelld:
I'm happy to wait for #631 to go in first.

gedankenstuecke:
Okay, I think I wrapped up all open work on #631 for now, let's get that in and then this one!

gedankenstuecke (approved):
Also just tested it all and that's now working perfectly, thanks @llewelld! 🚀

gedankenstuecke:
As feared, there's now some merge conflicts. @llewelld are you happy for me to try to resolve those? 😄

llewelld:
> @llewelld are you happy for me to try to resolve those? 😄

Thanks for the offer. I still have the code on my machine and some lingering context in my head, so I'd be keen still to do this.

gedankenstuecke:
That’s why I asked, I didn’t want to snatch away the work - but also don’t want you to feel obliged to do it! ☺️

llewelld:
I appreciate you asking! I wouldn't have wanted this to block progress, but equally I'd feel uncomfortable not finishing the task.

I've fixed the conflicts. I think I caught all of the Bootstrap 5 changes, but I may have missed some. Please do flag them up if you spot any.

gedankenstuecke (approved):
Looks good! 👍

llewelld:
Thank you @gedankenstuecke!

PR #643, opened 2024-01-16
fix button links in accordion on share

gedankenstuecke:
@GeorgiaHCA and I discovered that those buttons were still not linked and we'd like those to work for the demo with Autistica on Thursday! :D


GeorgiaHCA:
Fab yes go ahead and merge!

PR #645, opened 2024-01-19
Indicate that trigger-toggles worked

### LINKED ISSUE #606: Indiciate that page has been updated after content toggle changes
### Summary

This issue was flagged during user testing.

Make it clearer when triggering content toggles are changed that the stories have updated to include different content.

### What needs to be done?

There are potentially different ways that this could be addressed.

- [ ] Consider the options.
- [ ] Implement it.

Just as some initial ideas, this could be done using animations, transitions, a pop-up notification, a spinner, or something similar.

### Who can help?

* Design.
* Front-end development.

---

gedankenstuecke:
This fixes #606, by including a modal overlay of a "spinner" to indicate that the updated results are loaded. 

Technically, this is a bit smoke & mirrors, as it's completely unrelated to the actual loading, instead there's a quarter second period in which the overlay is shown before then reloading the website with the new results. 

While definitely not an elegant solution, I think for now this is a pragmatic approach, as the more dynamic implementation would mean a quite big chunk of work to refactor how the loading of the results happens!

### DIFF
--- a/server/apps/main/templates/main/experiences_page.html
+++ b/server/apps/main/templates/main/experiences_page.html
@@ -8,6 +8,16 @@
 
 {% block content %}
 
+<div class="modal fade" id="spinnerModal" tabindex="-1">
+  <div class="modal-dialog modal-dialog-centered justify-content-center">
+      <div class="text-center">
+        <div class="spinner-border" role="status">
+          <span class="visually-hidden">Refreshing...</span>
+        </div>
+    </div>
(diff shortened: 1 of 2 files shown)


helendduncan (approved):
On updating the toggles on the trigger list a quarter of a second spinner is overlayed on the screen before disappearing

gedankenstuecke:
Thanks @helendduncan for looking over it!

PR #648, opened 2024-01-31
fix sizing of animatiomn images

gedankenstuecke:
For some reason the original images of the animations shown on the front-page of AutSPACEs were unnecessary large to then be scaled down through CSS in the browser. 

This required nearly 10 MB of data to be transferred on load, which has now been shrunk down to about 1 MB instead. This will drastically improve load times and also avoid using up people's data allowances 🙂


helendduncan (approved):
There is no observable change in user experience with the resizing of the images

PR #649, opened 2024-02-02
[Ready for Review] Implement dark theme option

### LINKED ISSUE #604: [FEATURE]: Add a dark theme
### Feature for…

Frontend

### What is the problem you want to solve?

Suggestion made during user testing.

Some users prefer or find text more legible when using a dark theme (light text on a darker background). Many sites now support this and most browsers have support for switching themes based on a system-wide default. Users have therefore come to expect that an accessible site will support both a light and dark theme.

### How do you want to solve the problem?

Possible solutions:
 1. [ ] Provide a CSS dark theme and a switch on every page to toggle between the dark and light theme.
 2. [ ] Provide a CSS dark theme that toggles based on the system defult.
 3. [ ] Provide a CSS dark theme and a switch that can be toggled in the user profile.

### Other comments

Examples of sites with a dark theme:
1. [Github](https://github.com)
2. [Google](https://www.google.com)
3. [DuckDuckGo](https://duckduckgo.com)

Potentially useful reference material:
1. [The Ultimate Guide to Enabling Dark Mode Everywhere](https://www.howtogeek.com/449742/the-ultimate-guide-to-enabling-dark-mode-everywhere/)
2. [A Complete Guide to Dark Mode on the Web ](https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/)

helendduncan:
### Expected behaviour
On first load of the page detects users' system theme, and sets the theme to the corresponding option (light/dark)

Button added in Nav bar to toggle between light and dark modes. If selected, the selected theme is stored in Local Storage. 

Some re-defining classes in the css file to match the bootstrap defaults for more consistent switching from light to dark mode, and added a "grey" copy of the flowchart on the help page.

Things that perhaps need review/thought. 
- [ ] "Toggle Theme" should the wording in the nav bar be more explicit (switch light/dark mode) or in fact be a toggle?
- [ ] On the home page `server/apps/main/templates/main/home.html` there are inconsistencies with the background/foreground for the "About AutSPACEs" section when compared to the shared stories carousel section (what turns light/dark) but the text colour in the "About AutSPACEs" section doesn't look particularly contrasting when the background is dark. Needs someone with design veto to decide.

Closes #604


gedankenstuecke:
Thanks for all the work on this @helendduncan! 

I think typically dark mode is similar to "inverting" the colours, with mostly dark-ish backgrounds and light grey/off-white text on top of it. In that sense the second one is probably closer to that.

The top one might have the problem of being too low-contrast depending on accessiblity needs? 

tl;dr: if non-dark mode is already "white text on darker background" we might want to keep it that way in dark mode too :)

static/css/main.css
```diff
@@ -284,7 +285,7 @@ body {
  /*Platform Information*/
 
  #platform_info {
-   background: #FFFFFF;
+   background: var(--bs-body-bg);
```
gedankenstuecke, line 288: I just added this bit in, as otherwise that part remained white throughout!

gedankenstuecke (approved):
Looks great, thanks for all the hard work on this @helendduncan!

gedankenstuecke:
Also re-worded the navbar item a bit, happy to have it merged now!

PR #652, opened 2024-02-06
Clarify trigger defaults on profile page

### LINKED ISSUE #612: Make it clear triggering content preferences can be changed any time
### Summary

We currently allow people to select their preferences for what kinds of potentially triggering content they would like to view when they set up their profile. We should make it clear on the profile page that this is not a permanent choice and they will be able to change their preferences any time after they have registered. 

### What needs to be done?

- [ ] Include a clear sentence to say the preferences can be changed at any time, but that these will be used as a default
- [ ] Include clear instructions on how to change preferences

gedankenstuecke:
This closes #612 by providing a clearer labeling of what those toggle switchees do (and that they can be overwritten/changed)

### DIFF
--- a/server/apps/users/templates/users/profile.html
+++ b/server/apps/users/templates/users/profile.html
@@ -47,7 +47,7 @@
       {% endif %}
       {% if group.grouper == 2 %}
         {% define "Sensitive topics" as header %}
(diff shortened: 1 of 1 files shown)


helendduncan (approved):
Changes work as expected

gedankenstuecke:
Thanks @helendduncan!

PR #654, opened 2024-02-07
adds rss/atom feeds for public experiences

### LINKED ISSUE #610: [FEATURE]: Add RSS feeds
### Feature for…

Frontend and Backend

### What is the problem you want to solve?

Suggested during user testing.

RSS feeds would be useful: uses these to filter information and return to websites.

An RSS feed would allow users to access the content in a different way, supporting different workflows (e.g. where a user is interested to see all the material that's added, or is only interested in the latest changes).

A separate app is often used to read RSS feeds, which means the user can have greater control over how viewing is configured compared to using the website.

Adding an RSS feed might improve accessibility for some users.

### How do you want to solve the problem?

1. [ ] Provide a URL where the RSS feed can be accessed.
2. [ ] Add links to the URL on the main website.

### Other comments

RSS feeds can usually be accessed without a login. Would different RSS feeds therefore need to be made available in order to ensure different forms of triggering content are omitted?

Potentially useful references:
1. [RSS docs on W3Schools](https://www.w3schools.com/XML/xml_rss.asp).
2. [RSS spec](https://cyber.harvard.edu/rss/rss.html).
3. [Atom docs](https://validator.w3.org/feed/docs/atom.html).

gedankenstuecke:
Closes #610 

This PR adds both an Atom & RSS feed on the public experiences page, properly linked behind the scenes in the `head` of the website as appropriate to facilitate automatic discovery of it. 

It also includes a link to the feed on that page in order to highlight this flag beyond automated discovery. Furthermore, the feed links automatically mirror the current view including the trigger inclusion choices, so that people can subscribe to a feed that includes only those topics they are comfortable reading about.

### DIFF
--- a/server/apps/main/feeds.py
+++ b/server/apps/main/feeds.py
@@ -30,7 +30,7 @@ class PublicExperienceFeed(Feed):
 		return "AutSPACEs public experiences"
 
 	def link(self, obj):
-		return "https://example.com"
+		return reverse("main:public_experiences")
 
 	def description(self, obj):
 		joined_triggers = ", ".join(obj)
@@ -51,7 +51,7 @@ class PublicExperienceFeed(Feed):
 	def items(self, obj):
 		experiences = PublicExperience.objects.filter(moderation_status="approved")
 		experiences = expand_filter(experiences, obj)
-		experiences = experiences.order_by("-created_at")[:5]
+		experiences = experiences.order_by("-created_at")[:10]
 		return experiences
 
 class PublicExperienceAtomFeed(PublicExperienceFeed):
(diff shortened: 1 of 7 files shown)


gedankenstuecke:
Okay, I just added the tests as well to keep our coverage up, @helendduncan I think this would also be ready for review now! 🚀

helendduncan (approved):
I'm not familiar with RSS feeds, but it looks like it does what's required - only comment/query would be if it would make sense to have the RSS under the stories, vs above the stories?

gedankenstuecke:
I think that's a good idea, especially as the website itself should in the metadata announce the presence of an RSS feed, so that it's accessible by just linking to the view experiences page!

PR #658, opened 2024-02-14
Bump Django/Python versions

gedankenstuecke:
Django 5.x has [lots of accessibility improvements](https://www.djangoproject.com/weblog/2024/feb/10/django-accessibility-in-2023-and-beyond/) which would be useful to have - and also generally good to not fall too behind in terms of versions. 

This tries to update us to both Django 5.x as well as Python 3.12


gedankenstuecke:
Merging this in to get started on the last usability improvements uncovered by the second round of user tests!

PR #681, opened 2024-05-23
[WIP] Make docker for production

gedankenstuecke:
This is a work in progress, requiring adapting the image to modern docker deployments

this includes the static handling


gedankenstuecke:
Okay, this PR does a number of things to allow an actual deployment as a _web app_ on Azure: 

## Custom Dockerfile

 I've made a custom `Dockerfile` in the root for this, which is based on bits of the existing file, but as we can't/won't be using `docker swarm` we needed to adapt it to the Azure environment. It now still takes all the settings etc. as before from the production setup

## Use Whitenoise
The previous setup was expecting to deploy with an extra webserver in addition to gunicorn for managing the static files. While possible, this is a pain to manage on Azure. 

Instead, I've added in `whitenoise`, which is a simple library to handle serving of static files for smallish web apps like ours. 

As part of this I also had to adapt and 'fix' many of our `static` falls, as they included a wrong leading `/`. 

## CSP exceptions
In development we had already included the exceptions to load 3rd-party CSS/JS for Bootstrap et al. In production settings those were missing, I've added those. 

## Concurrent tasks
The existing config wanted to spawn 4 `gunicorn` workers in parallel, our low-cost Azure deployment was too small for that, so I reduced it to 2 instead.

gedankenstuecke:
Extra notes: For the production deployment there's two main settings that are important to enable it to work (which aren't necessary in `development`): 

1. `DJANGO_ENV` as the environment variable needs to be set to `production`
2. `DOMAIN_NAME` needs to be set to the actual domain it's being deployed to, for safety reasons the server is setup to otherwise only accept `localhost`
3. `WEBSITES_PORT` - specific to Azure, to tell it which port to connect to.

gedankenstuecke:
I've added in a GH workflow to automatically push the image to Docker Hub when merging things into `main`.

PR #687, opened 2024-06-06
Participant information sheet

Davsarper:
latest and approved participant information sheet


gedankenstuecke (approved):
Thanks @Davsarper!

